PR #109, opened 2026-06-07
fix(daemon): reconcile orphaned workflow instances at startup

orlandoburli:
## Summary

At daemon startup, automatically mark any `workflow_instances` left in the `running` state from a previously-killed process as `interrupted`. This prevents tasks from appearing stuck and allows them to re-dispatch naturally on the next poll.

**Problem:** When the daemon crashes or is forcefully killed, any running workflow instances are left stranded in the database. On restart, the daemon owns no in-flight runs, so these orphans are never resolved. Tasks appear stuck (showing "running" far longer than daemon uptime) and won't re-dispatch because the dispatcher sees an existing running instance and drops matching routes.

**Solution:** At `Dispatcher.Start()`, reconcile orphaned running instances by marking them `interrupted`. This is analogous to `ReconcileOrphanExecutions` (which handles task_executions). The next poll then dispatches fresh instances naturally.

## Implementation

- Added reconciliation call in `Dispatcher.Start()` after `ReconcileOrphanExecutions`
- `ReconcileOrphanWorkflowInstances` marks only `running` instances as `interrupted`
- `approval_waiting` instances are intentionally untouched (rehydrated separately via `rehydrateParkedApprovals`)
- Added test coverage for the reconciliation logic

## Test Plan

- [x] Unit tests pass: TestReconcileOrphanWorkflowInstances_Extended verifies running → interrupted, approval_waiting untouched
- [x] All integration tests pass (go test ./...)
- [x] Build succeeds (make build)

### DIFF
--- a/src/internal/daemon/dispatcher.go
+++ b/src/internal/daemon/dispatcher.go
@@ -291,6 +291,17 @@ func (d *Dispatcher) Start(ctx context.Context, wg *sync.WaitGroup) {
 		} else if n > 0 {
 			aplog.Info("reconciled %d orphaned running execution(s) from a previous run", n)
 		}
+
+		// Mark any workflow instances left in the 'running' state as interrupted.
+		// These are orphans from a previously-killed daemon that would otherwise
+		// cause tasks to appear stuck (running for longer than daemon uptime).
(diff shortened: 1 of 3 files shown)


orlandoburli:
Merged locally via squash. Work is on main.

PR #129, opened 2026-06-09
feat(tokens): store cache_creation/cache_read tokens separately + capture Cursor

orlandoburli:
## What

Token usage previously **folded cache tokens into the input side**, so the dashboard could show input/output/total/cost but not the cache breakdown — even though cache writes and cache reads are priced very differently. This stores the two cache token types as their own columns and surfaces them in the Task Detail.

It also **fixes Cursor token capture**: the Cursor agent CLI emits a `result` event with the same shape as Claude but **camelCase** usage keys (`inputTokens`/`outputTokens`/`cacheWriteTokens`/`cacheReadTokens`), so its tokens were silently recording as zero. The parser now accepts both namings. Verified against live `cursor-agent -p --output-format stream-json` output.

## Design

`InputTokens` **stays the full billed input** (cache folded in), so every existing total/in/out display and aggregate `SUM` is unchanged. The cache breakdown rides alongside as `CacheCreationTokens` / `CacheReadTokens`. Pure uncached input is therefore `input − creation − read`. Fully additive — zero behavior change for existing data; legacy rows read 0 for the new columns.

## Changes

- **model.Usage** — `CacheCreationTokens` / `CacheReadTokens`.
- **runner/execution/cli.go** — parser accepts Claude (snake_case) and Cursor (camelCase) usage; records the cache split; `applyUsage` helper.
- **db** — `cache_creation_tokens` + `cache_read_tokens` columns on `task_executions` and `step_runs`; threaded through `Execution`/`StepRun` structs, INSERT/UPDATE/SELECT, and the step/instance usage rollups.
- **daemon/workflow.go, workflow/engine.go, workflow_ipc.go** — threaded through per-attempt → per-step → per-instance summing and the IPC step payload.
- **dashboard** — Task Detail step view shows a `Cache: N write / N read` line when present.

## Not included

`opencode-cli` still reports no tokens — it runs without a structured output format and its JSON mode surfaces no usage event. Capturing it would need an invocation change + a new event shape, out of scope here.

## Tests

- `TestAccumulateStreamUsage_Full` extended with cache-breakdown assertions.
- New `TestAccumulateStreamUsage_CursorCamelCase` (live-verified Cursor shape).
- `StepRun` persistence test round-trips the new columns.
- `go build ./...`, `go vet ./...`, and full `go test ./...` all green.

### DIFF
--- a/src/internal/cli/instances.go
+++ b/src/internal/cli/instances.go
@@ -138,6 +138,16 @@ func showInstance(id string, asJSON bool) error {
 				s.Duration,
 				stateColor(s.State).Render(state),
 			)
+			if s.TotalTokens > 0 {
+				usage := fmt.Sprintf("%d in / %d out / %d total", s.InputTokens, s.OutputTokens, s.TotalTokens)
+				if s.CacheCreationTokens > 0 || s.CacheReadTokens > 0 {
+					usage += fmt.Sprintf("  ·  cache %d write / %d read", s.CacheCreationTokens, s.CacheReadTokens)
+				}
+				if s.CostUSD > 0 {
+					usage += fmt.Sprintf("  ·  $%.5f", s.CostUSD)
+				}
+				fmt.Printf("       %s\n", instMuted.Render(usage))
+			}
 		}
 	}
 	printCIPolls(detail.CIPolls)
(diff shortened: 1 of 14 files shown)


orlandoburli:
Added cache-breakdown visibility at two more surfaces (commit 6c0577f):

- **Dashboard Task Detail header** — `Cache: N write / N read` rolled up across the task's workflow instances (under Tokens); the Workflow Instances list one-liner also shows `· N cache`.
- **`apiary instances <id>` CLI** — each step now prints a usage line (tokens · cache write/read · cost). The CLI previously showed no token info per step at all.

Token visibility now spans: per-step detail, task/instance header, and the CLI.

PR #253, opened 2026-07-24
security: strip host env vars via allow-list before spawning agent subprocesses (SEC-07)

orlandoburli:
## Summary

- Agent subprocesses previously inherited the **full daemon environment** via `os.Environ()`, silently exposing every ambient secret (cloud credentials, API tokens, webhook secrets, database URLs) to agent-run code.
- Introduces `hostEnv()` in a new `execution/env.go` file: an explicit allow-list that keeps only well-known safe variables (`PATH`, `HOME`, locale, SSH agent socket, XDG dirs, terminal vars) and **strips everything else**.
- Credentials agents legitimately need (`GITHUB_TOKEN`, `ANTHROPIC_API_KEY`, …) must now be declared in `agent.env` / `workflow.env` / `step.env` config and arrive via `req.Env`, which is overlaid on top of the filtered environment — as they already were for `GITHUB_TOKEN` via `agentIdentityEnv()`.
- Updates the `agentIdentityEnv()` doc comment to document the preferred **per-repo fine-grained token** pattern via `agent.source_token` rather than a broad host `GITHUB_TOKEN`.
- Adds `env_test.go` with four table-driven tests covering secret stripping, locale pass-through, full allow-list coverage, and unlisted-key exclusion.

## Test plan

- [x] `go test ./...` — all 23 packages pass
- [x] `go test ./internal/runner/execution/...` — 4 new env tests pass
- [ ] Verify that a deployed agent can still `git push` and call `gh` (needs `GH_TOKEN` / `GITHUB_TOKEN` in `agent.env` or via `source_token`)
- [ ] Confirm no ambient secrets appear in the subprocess environment in a real run

Closes #230

### DIFF
--- a/.claude/skills/gitnexus/gitnexus-cli/SKILL.md
+++ b/.claude/skills/gitnexus/gitnexus-cli/SKILL.md
@@ -5,16 +5,14 @@ description: "Use when the user needs to run GitNexus CLI commands like analyze/
 
 # GitNexus CLI Commands
 
-Commands below use `node .gitnexus/run.cjs <command>` — the project-local runner `gitnexus analyze` drops next to the index. It auto-selects an available runner at call time (global `gitnexus`, else `pnpm dlx`, else `npx`), so no package-manager assumption and no global install is required.
-
-> **Not analyzed yet, or `node .gitnexus/run.cjs` reports `Cannot find module`** (the gitignored runner is absent — e.g. a fresh clone or `git clean`)? (Re)generate it with `npx gitnexus analyze` from the project root. On **npm 11.x**, if `npx` crashes during install (`node.target is null`), install once with `npm i -g gitnexus` (then `gitnexus analyze`) or use `pnpm --allow-build=@ladybugdb/core --allow-build=gitnexus --allow-build=tree-sitter dlx gitnexus@latest analyze`. See [#1939](https://github.com/abhigyanpatwari/GitNexus/issues/1939).
+All commands work via `npx` — no global install required.
 
 ## Commands
 
 ### analyze — Build or refresh the index
 
 ```bash
-node .gitnexus/run.cjs analyze
+npx gitnexus analyze
 ```
 
 Run from the project root. This parses all source files, builds the knowledge graph, writes it to `.gitnexus/`, and generates CLAUDE.md / AGENTS.md context files.
@@ -24,14 +22,13 @@ Run from the project root. This parses all source files, builds the knowledge gr
 | `--force`      | Force full re-index even if up to date                           |
 | `--embeddings` | Enable embedding generation for semantic search (off by default) |
 | `--drop-embeddings` | Drop existing embeddings on rebuild. By default, an `analyze` without `--embeddings` preserves them. |
-| `--pdg` | Build the program-dependence layers used by `explain` and `pdg_query` (taint, CDG, and REACHING_DEF). |
 
-**When to run:** First time in a project, after major code changes, or when `gitnexus://repo/{name}/context` reports the index is stale. In Claude Code, a PostToolUse hook detects staleness after `git commit` and `git merge` and notifies the agent to run `analyze` — the hook does not run analyze itself, to avoid blocking the agent for up to 120s and risking KuzuDB corruption on timeout.
+**When to run:** First time in a project, after major code changes, or when `gitnexus://repo/{name}/context` reports the index is stale. In Claude Code, a PostToolUse hook runs `analyze` automatically after `git commit` and `git merge`, preserving embeddings if previously generated.
 
 ### status — Check index freshness
 
 ```bash
-node .gitnexus/run.cjs status
+npx gitnexus status
 ```
 
 Shows whether the current repo has a GitNexus index, when it was last updated, and symbol/relationship counts. Use this to check if re-indexing is needed.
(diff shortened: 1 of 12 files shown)


orlandoburli:
## Code Review — SEC-07 env allow-list

**Verdict: LGTM.** One non-blocking follow-up noted below.

### Core implementation ✅

`env.go` / `cli.go` changes are correct and complete:

- `hostEnv()` properly filters `os.Environ()` to an explicit allow-list. The design rationale comment is thorough.
- `cli.go:107-110` assigns `hostEnv()` first, then overlays `req.Env` — correct ordering ensures operator-declared credentials win without leaking ambient ones.
- Allow-list is well-calibrated: `PATH`, `HOME`, locale, SSH agent socket, XDG dirs, terminal vars, `CI`; no credential namespaces.
- `agentIdentityEnv()` doc comment now documents the `agent.source_token` per-repo fine-grained token pattern.

### Tests ✅

Four tests in `env_test.go` cover secret stripping, locale prefix pass-through, full allow-list iteration, and unlisted-key exclusion. `containsKey` uses `key + "="` prefix matching — correctly avoids false positives.

### Existing plugin subprocess ✅

`plugin/client.go` already had its own allow-list via the `environment()` method (`PATH`, `HOME`, `TMPDIR`, `TEMP`, `LANG`, `LC_ALL`, `TZ` + `Security.SecretEnv`). Not in scope here and already secure.

### Follow-up note (non-blocking) ⚠️

`daemon/notify.go:67` still passes `os.Environ()` to operator-configured notification command subprocesses:

```go
cmd.Env = append(os.Environ(),
    "APIARY_TASK_ID="+ev.TaskID, ...
)
```

Since this is operator-configured (not agent-controlled), the exfiltration risk is lower than for agent subprocesses. Worth a follow-up issue to apply `hostEnv()` here too. Not a blocker for this PR given the stated scope.

orlandoburli:
## Code Review Council — PR #253 (SEC-07: env allow-list before spawning agent subprocesses)

**VERDICT: PASS** (4/5 APPROVE, no veto-lens rejection)

### Correctness Hawk (veto) — APPROVE
`hostEnv()` correctly filters `os.Environ()` and the overlay order is right: `cmd.Env = hostEnv()` then `req.Env` appended after, so — since `os/exec` honours the *last* value for duplicate keys — a configured `GITHUB_TOKEN`/`PATH` wins over any host value. `strings.Cut` handles `=` in values correctly; malformed (no `=`) entries are safely skipped. The only agent-subprocess spawn is `CliRunner.Run` (`cli.go`) and it is covered; `api.go` is HTTP, `cli_mcp.go` only assembles config (MCP children inherit the already-filtered CLI env). Tests are sound and cover strip/pass-through/full-allow-list/exclusion.

### Security Skeptic (veto) — APPROVE
Default-deny allow-list is the correct design and directly closes the ambient-secret exfiltration hole (#230). No bypass in the matcher. One residual to note (not blocking): `SSH_AUTH_SOCK` is passed through, so an agent can still authenticate via the host SSH agent as the daemon's SSH identity — this partially undercuts the "per-repo fine-grained token" goal the doc comment promotes, but it is required for `git push` over SSH and is a reasonable, documented tradeoff.

### Conventions Stickler — REJECT
Scope creep. A security PR bundles unrelated regeneration of `.claude/skills/gitnexus/*`, `AGENTS.md`, and `CLAUDE.md` (symbol counts 6306→10279, removed tool docs) — the bulk of the 190 deletions. These are unrelated to SEC-07 and, per project practice, `AGENTS.md`/`CLAUDE.md` should not be regenerated into tracked commits. The Go change itself follows conventions (new file, table-driven tests, thorough doc comments). Recommend splitting the doc churn out.

Minimal, readable core change; a single well-scoped helper with a clear map + prefix list. No over-engineering.

### Pragmatist — APPROVE
Ships real security value at low, reversible risk. Watch-outs: `HTTP(S)_PROXY`/`NO_PROXY` and git-signing vars (`GNUPGHOME`, `GPG_TTY`) are not allow-listed and may break agents in proxied or commit-signing setups — surface in release notes as "declare via agent.env if needed." The two real-run test-plan items (agent can still `git push`/`gh`; no ambient secrets in a live run) remain unchecked and should be verified before rollout.

**Decision:** No Correctness/Security veto; 4/5 APPROVE ≥ 3 → **PASS**. Please split the gitnexus/CLAUDE.md/AGENTS.md doc changes into a separate PR before merge.

orlandoburli:
The implementation is correct and well-structured. A few notes:

**What's done well:**
- Deny-by-default approach is exactly right — explicit allow-list rather than a block-list means new secrets can never accidentally leak through.
- `req.Env` overlay happens *after* `hostEnv()`, so credentials declared in `agent.env`/`workflow.env`/`source_token` still reach the subprocess without needing to be on the allow-list.
- `strings.Cut` is the correct way to split `KEY=value` pairs (handles values that themselves contain `=`).
- `t.Setenv` in tests is idiomatic — it restores the original value automatically, so tests don't bleed into each other.
- `containsKey` correctly uses `key + "="` as a prefix, which avoids false-positives (e.g., searching for `PATH` will not match `PATH_EXTRA=...`).

**One thing to keep in mind (not a blocker):**
`SSH_AUTH_SOCK` on the allow-list gives agent subprocesses access to the SSH agent, which lets them authenticate as the host user for any SSH-accessible repository — including ones not related to the current workflow. This is a documented and reasonable trade-off for agents that push over SSH, but operators running in high-threat environments may want to disable the SSH agent socket before invoking apiary (e.g., `unset SSH_AUTH_SOCK`) or use HTTPS + `source_token` exclusively. Worth a mention in the operator docs.

**Minor:**
`UID` is typically a bash internal not exported to the environment in most shells, so it will rarely appear in `os.Environ()`. Harmless to keep in the allow-list, but can be removed if desired to avoid implying it is commonly present.

Overall: this closes the exfiltration surface area for SEC-07 correctly. Approved.

PR #254, opened 2026-07-24
security: sandbox agent execution to contain prompt injection (SEC-09)

orlandoburli:
## Summary

- **Minimal env allowlist** (`runner/execution/cli.go`): replaces the blanket `os.Environ()` inheritance with `filteredHostEnv()`, which only passes safe host variables (PATH, HOME, TMPDIR, etc.) to agent subprocesses. API keys, AWS credentials, and other daemon secrets are no longer leaked.

- **Docker sandbox wrapper** (`runner/execution/cli.go`): new `cliSandbox` type wraps agent subprocesses in a `docker run --rm` with a restricted user (default: `nobody`), network policy (default: `none`), and a bind-mounted working directory. Per-task credentials flow via `--env` flags, not host inheritance. Enabled via `sandbox:` in runner config.

- **Least-privilege OpenCode permissions** (`daemon/dispatcher.go`): `writeOpencodeAgent` and `registerAgentInConfig` now default to `bash: deny`, `edit: deny`, `webfetch: deny`. Agents that need shell/write access opt in via `permissions: {bash: allow, edit: allow}` in their `AgentConfig`.

- **New config types** (`config/config.go`): `SandboxConfig` (runner-level) and `Permissions map[string]string` (agent-level).

## Usage example

```yaml
runners:
  - id: opencode-sandboxed
    type: cli
    provider: opencode
    sandbox:
      image: "ghcr.io/example/opencode-agent:latest"
      user: "nobody"
      network: "none"
    config:
      command: opencode

agents:
  - id: engineer
    runner: opencode-sandboxed
    # opt in to bash + edit only when truly needed:
    permissions:
      bash: allow
      edit: allow
```

## Test plan

- [x] 8 new unit tests in `sandbox_test.go` cover: secret exclusion, allowlist preservation, overlay precedence, docker arg construction, default user/network, configure parsing, missing-image error, nil-overlay safety
- [x] Full test suite passes (`go test ./...` — all 24 packages)
- [x] `go build ./...` clean

Closes #232

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,25 +1,24 @@
 <!-- gitnexus:start -->
 # GitNexus — Code Intelligence
 
-This project is indexed by GitNexus as **apiary** (6306 symbols, 22913 relationships, 300 execution flows). Use the GitNexus MCP tools to understand code, assess impact, and navigate safely.
+This project is indexed by GitNexus as **apiary** (10303 symbols, 23448 relationships, 300 execution flows). Use the GitNexus MCP tools to understand code, assess impact, and navigate safely.
 
-> Index stale? Run `node .gitnexus/run.cjs analyze` from the project root — it auto-selects an available runner. No `.gitnexus/run.cjs` yet? `npx gitnexus analyze` (npm 11 crash → `npm i -g gitnexus`; #1939).
+> If any GitNexus tool warns the index is stale, run `npx gitnexus analyze` in terminal first.
 
 ## Always Do
 
-- **MUST run impact analysis before editing any symbol.** Before modifying a function, class, or method, run `impact({target: "symbolName", direction: "upstream"})` and report the blast radius (direct callers, affected processes, risk level) to the user.
-- **MUST run `detect_changes()` before committing** to verify your changes only affect expected symbols and execution flows. For regression review, compare against the default branch: `detect_changes({scope: "compare", base_ref: "main"})`.
+- **MUST run impact analysis before editing any symbol.** Before modifying a function, class, or method, run `gitnexus_impact({target: "symbolName", direction: "upstream"})` and report the blast radius (direct callers, affected processes, risk level) to the user.
+- **MUST run `gitnexus_detect_changes()` before committing** to verify your changes only affect expected symbols and execution flows.
 - **MUST warn the user** if impact analysis returns HIGH or CRITICAL risk before proceeding with edits.
-- When exploring unfamiliar code, use `query({search_query: "concept"})` to find execution flows instead of grepping. It returns process-grouped results ranked by relevance.
-- When you need full context on a specific symbol — callers, callees, which execution flows it participates in — use `context({name: "symbolName"})`.
-- For security review, `explain({target: "fileOrSymbol"})` lists taint findings (source→sink flows; needs `analyze --pdg`).
+- When exploring unfamiliar code, use `gitnexus_query({query: "concept"})` to find execution flows instead of grepping. It returns process-grouped results ranked by relevance.
+- When you need full context on a specific symbol — callers, callees, which execution flows it participates in — use `gitnexus_context({name: "symbolName"})`.
 
 ## Never Do
 
-- NEVER edit a function, class, or method without first running `impact` on it.
+- NEVER edit a function, class, or method without first running `gitnexus_impact` on it.
 - NEVER ignore HIGH or CRITICAL risk warnings from impact analysis.
-- NEVER rename symbols with find-and-replace — use `rename` which understands the call graph.
-- NEVER commit changes without running `detect_changes()` to check affected scope.
+- NEVER rename symbols with find-and-replace — use `gitnexus_rename` which understands the call graph.
+- NEVER commit changes without running `gitnexus_detect_changes()` to check affected scope.
 
 ## Resources
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,25 +1,24 @@
 <!-- gitnexus:start -->
 # GitNexus — Code Intelligence
 
-This project is indexed by GitNexus as **apiary** (6306 symbols, 22913 relationships, 300 execution flows). Use the GitNexus MCP tools to understand code, assess impact, and navigate safely.
+This project is indexed by GitNexus as **apiary** (10303 symbols, 23448 relationships, 300 execution flows). Use the GitNexus MCP tools to understand code, assess impact, and navigate safely.
 
-> Index stale? Run `node .gitnexus/run.cjs analyze` from the project root — it auto-selects an available runner. No `.gitnexus/run.cjs` yet? `npx gitnexus analyze` (npm 11 crash → `npm i -g gitnexus`; #1939).
+> If any GitNexus tool warns the index is stale, run `npx gitnexus analyze` in terminal first.
 
 ## Always Do
 
-- **MUST run impact analysis before editing any symbol.** Before modifying a function, class, or method, run `impact({target: "symbolName", direction: "upstream"})` and report the blast radius (direct callers, affected processes, risk level) to the user.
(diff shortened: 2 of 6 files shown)


orlandoburli (commented):
## Code Review — SEC-09 Sandbox

Overall the design is solid and addresses the core issue. Three findings worth addressing before merge, one of which is a silent breaking change.

---

### 1. `filteredHostEnv` silently breaks non-sandbox runners (medium)

**File:** `src/internal/runner/execution/cli.go`, non-sandbox branch of `Run()`

```go
// before
cmd.Env = os.Environ()
for k, v := range req.Env { ... }

// after (non-sandbox path)
cmd.Env = filteredHostEnv(req.Env)
```

The allowlist is applied to **all** runners, not just sandboxed ones. Any existing deployment that relies on host env vars not in the allowlist — `GOPATH`, `GIT_AUTHOR_NAME`, `XDG_CONFIG_HOME`, npm/pip config paths, etc. — will silently break without enabling sandbox at all.

The allowlist-only approach is correct security posture for sandboxed runners. For non-sandbox runners (where the operator already trusts the environment) the old `os.Environ()` + overlay was fine, and this change removes it without a config option or migration note.

**Suggested fix:** keep `os.Environ()` + overlay in the non-sandbox branch; restrict the allowlist to the sandbox path only. The security gain of `filteredHostEnv` in non-sandbox mode is marginal (the subprocess runs under the daemon user anyway).

```go
if r.sandbox != nil {
    binary, sandboxArgv := r.sandbox.wrapCommand(...)
    cmd = exec.CommandContext(ctx, binary, sandboxArgv...)
    cmd.Env = filteredHostEnv(nil)            // docker host process only needs the allowlist
} else {
    cmd = exec.CommandContext(ctx, r.command, argv...)
    cmd.Dir = req.WorkingDir
    cmd.Env = os.Environ()                    // non-sandbox: inherit full daemon env
    for k, v := range req.Env {
        cmd.Env = append(cmd.Env, k+"="+v)
    }
}
```

---

### 2. Container filesystem is writable (low)

**File:** `src/internal/runner/execution/cli.go`, `wrapCommand()`

The `docker run` invocation has no `--read-only`, `--cap-drop`, or `--security-opt no-new-privileges`. A sandboxed agent running as `nobody` can still write to `/tmp`, `/var`, etc., install suid helpers, or accumulate state across tasks if Docker layer caching is involved.

Suggested additions before the image argument:
```go
dockerArgs = append(dockerArgs,
    "--read-only",
    "--tmpfs", "/tmp:rw,noexec,nosuid,size=64m",
    "--cap-drop", "all",
    "--security-opt", "no-new-privileges",
)
```

These should be defaults that operators can override via `extra_args`, not the other way around. Also worth considering `--memory 512m --pids-limit 64` to prevent resource exhaustion from runaway injected commands.

---

### 3. Nit: `extraArgs` ordering allows privilege escalation

**File:** `src/internal/runner/execution/cli.go`, `wrapCommand()` line ordering

```go
dockerArgs = append(dockerArgs, s.extraArgs...)    // operator extra args
dockerArgs = append(dockerArgs, s.image, binary)   // then image
```

This is correct — `extraArgs` are docker options, not container commands, so ordering before the image is right. But there's no validation that extra_args entries don't include `--user root` or `--privileged`, which would defeat the sandbox. This is operator-controlled config (not user-controlled), so it's acceptable, but a comment noting the trust boundary would help: `// extraArgs are trusted operator overrides; only use for bind-mounts or named networks`.

---

### Minor observations

- `TestFilteredHostEnv_OverlayTakesPrecedence` verifies the overlay value appears but doesn't assert that the allowlisted version of the same key is **absent**. If both `HOME=/home/daemon` and `HOME=/home/task` appear, exec will use whichever is last (overlay wins in practice), but the test doesn't catch a regression that emits both.
- The PR description's usage example is clear. Shipping it as a comment or snippet in `AGENTS.md` would help operators discover the sandbox option without reading the struct.

---

**Verdict:** Approve pending fix for finding #1, which is the only correctness issue. Findings #2/#3 are hardening improvements — acceptable as follow-up if you want to keep scope tight.

orlandoburli:
## Review Council Verdict — PR #254 (SEC-09 sandbox / env allowlist)

**Decision: REJECTED** — Correctness Hawk and Security Skeptic both veto (either alone blocks; decision rule: any Correctness/Security reject => REJECTED). Tally: 1 APPROVE / 4 REJECT.

---

### 🦅 Correctness Hawk — REJECT (veto) — ROLLOUT BLOCKER
`filteredHostEnv` strips **the LLM provider credentials the CLI agents actually need to run.** The allowlist (PATH/HOME/TMPDIR/…) contains no `ANTHROPIC_API_KEY`, `OPENAI_API_KEY`, `OPENCODE_API_KEY`, or Cursor token, and these keys are **not** in `req.Env` either — `stepEnv`/`agentIdentityEnv` (src/internal/daemon/workflow.go:879-918) only inject git identity + `GITHUB_TOKEN`/`GH_TOKEN` + memory dir. The CLI adapters (claude/opencode/cursor) authenticate purely from the daemon's inherited host env; `cli.go` has no `api_key` injection (only the separate API adapter in `api.go` does). Before this PR: `cmd.Env = os.Environ()` carried the provider key through. After: `cmd.Env = filteredHostEnv(req.Env)` drops it. Net effect: **every CLI agent loses provider auth on upgrade → 100% of agent runs fail.** This is exactly the kind of silent-break a rollout gate exists to catch. The PR conflates "daemon secrets to hide from the agent" with "the one credential the agent legitimately requires."

### 🔒 Security Skeptic — REJECT (veto)
1. **Per-task secrets leaked into the host process table.** `wrapCommand` passes credentials as `--env KEY=VALUE` on the `docker run` argv (cli.go). Command-line args are world-readable via `/proc/<pid>/cmdline` / `ps aux`, so `GITHUB_TOKEN` (and any scoped token) is exposed to every local user/process for the container's lifetime. Use `--env KEY` (value pulled from docker's own env) or `--env-file` instead.
2. **Bind mount + `--user nobody` will fail writes** on a daemon-owned `workDir`, and `-v workDir:workDir` re-exposes the full working tree; combined with the ineffective isolation this undercuts the stated containment goal.
3. Env-allowlist intent is sound, but shipping it without the provider key (see Correctness) means operators will be pushed to re-add secrets via `agent.Env`, landing them right back in the argv leak from (1).

### 📐 Conventions Stickler — REJECT
CLAUDE.md mandates `impact` before editing symbols and `detect_changes` before commit; the PR touches high-fan-in symbols (`writeOpencodeAgent`, `registerAgentInConfig`, `CliRunner.Run`, `Configure`) with no impact/blast-radius evidence. The PR also silently rewrites AGENTS.md/CLAUDE.md GitNexus docs (unrelated to SEC-09) — memory rule is to never dirty those tracked files (`--skip-agents-md`). Scope creep in a security PR.

### ✂️ Simplicity Advocate — REJECT
Two separate least-privilege permission maps are duplicated verbatim in `writeOpencodeAgent` and `registerAgentInConfig` (drift risk); factor into one helper. Otherwise the shape is reasonable.

### ⚙️ Pragmatist — APPROVE (conditional)
Direction is right and tests are decent. But default `network: "none"` makes the sandbox non-functional out of the box (agent can't reach the LLM API), and the env-allowlist regression means this cannot ship as-is. Approve the *approach*, not this diff.

---

**Required before re-review:**
1. Add provider credential(s) to the pass-through path (allowlist the configured provider key, or inject it explicitly into `req.Env`) — verify a real agent run authenticates.
2. Pass container secrets via `--env KEY` / `--env-file`, never `--env KEY=VALUE` on argv.
3. Provide `impact`/`detect_changes` output for the touched symbols; revert the unrelated AGENTS.md/CLAUDE.md edits.
4. De-duplicate the permission-default map.

*Automated council review — not an approval; do not merge.*

orlandoburli:
Superseded by rework issue #289 (council rejected this PR with specific defects; rework will open a fresh PR through the review gate). Closing to avoid duplicate PRs.

PR #255, opened 2026-07-24
feat(plugin): enforce manifest checksum + network sandbox (SEC-10)

orlandoburli:
## Summary

- **Unsigned plugins refused**: `Manifest.Checksum` (SHA-256 hex of the executable) is now mandatory. `Load()` rejects any plugin whose checksum field is missing or whose computed digest differs from the declared one. `NewClient()` re-verifies to close the TOCTOU window between discovery and first use.
- **Network sandbox on Linux**: when `security.network` is `false` (the default), the plugin subprocess is launched inside an unprivileged `user+network` namespace (`CLONE_NEWNET | CLONE_NEWUSER`). The namespace has no external interfaces — only loopback — so the plugin cannot make outbound connections. When `network: true` the subprocess runs without a namespace. Other platforms receive a no-op stub (`sandbox_other.go`); the restricted env-var allowlist from `environment()` still applies.
- **Path validation**: `security.read_paths` / `security.write_paths` entries are now validated at load time: empty entries and `..`-escape paths are rejected. Runtime enforcement (Linux landlock / seccomp) is left for a follow-up once the sandbox boundary is proven.
- **`ComputeChecksum` helper**: exported function for plugin authors to generate the digest before publishing their manifest.

## Test plan

- [x] `TestUnsignedPluginRejected` — manifest with no checksum fails `Load`
- [x] `TestChecksumMismatchRejected` — wrong digest fails `Load`
- [x] `TestChecksumVerifiedAtNewClient` — binary replaced between `Load` and `NewClient` fails `NewClient`
- [x] `TestSecurityPathValidation` — `..` escape and empty path entries rejected; valid paths accepted
- [x] `TestNetworkSandboxIsolatesPlugin` (Linux only) — plugin sees 0 non-loopback interfaces inside sandbox
- [x] `TestNetworkSandboxAllowedWhenDeclared` (Linux only) — `network: true` plugin invokes successfully
- [x] All pre-existing tests pass unchanged (checksum is now computed by `writeTestPlugin` automatically)
- [x] `GOOS=linux GOARCH=amd64 go build` succeeds

Closes #233

### DIFF
--- a/.claude/skills/gitnexus/gitnexus-cli/SKILL.md
+++ b/.claude/skills/gitnexus/gitnexus-cli/SKILL.md
@@ -5,16 +5,14 @@ description: "Use when the user needs to run GitNexus CLI commands like analyze/
 
 # GitNexus CLI Commands
 
-Commands below use `node .gitnexus/run.cjs <command>` — the project-local runner `gitnexus analyze` drops next to the index. It auto-selects an available runner at call time (global `gitnexus`, else `pnpm dlx`, else `npx`), so no package-manager assumption and no global install is required.
-
-> **Not analyzed yet, or `node .gitnexus/run.cjs` reports `Cannot find module`** (the gitignored runner is absent — e.g. a fresh clone or `git clean`)? (Re)generate it with `npx gitnexus analyze` from the project root. On **npm 11.x**, if `npx` crashes during install (`node.target is null`), install once with `npm i -g gitnexus` (then `gitnexus analyze`) or use `pnpm --allow-build=@ladybugdb/core --allow-build=gitnexus --allow-build=tree-sitter dlx gitnexus@latest analyze`. See [#1939](https://github.com/abhigyanpatwari/GitNexus/issues/1939).
+All commands work via `npx` — no global install required.
 
 ## Commands
 
 ### analyze — Build or refresh the index
 
 ```bash
-node .gitnexus/run.cjs analyze
+npx gitnexus analyze
 ```
 
 Run from the project root. This parses all source files, builds the knowledge graph, writes it to `.gitnexus/`, and generates CLAUDE.md / AGENTS.md context files.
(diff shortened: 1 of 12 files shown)


orlandoburli:
## 🐝 Apiary Review Council — PR #255 (SEC-10)

**Verdict: ❌ REJECTED** (Correctness Hawk veto)

The checksum + path-validation work is solid, but the network-sandbox runtime path and the unrelated bundled doc reverts block this.

### Lenses

**🦅 Correctness Hawk — ❌ REJECT (veto)**
`applySandbox` (sandbox_linux.go) unconditionally sets `CLONE_NEWUSER | CLONE_NEWNET` for every invoke when `network:false` — which is the *default*. On any host where unprivileged user namespaces are disabled (hardened kernels with `kernel.unprivileged_userns_clone=0`, `max_user_namespaces=0`, and many container runtimes) `cmd.Start()` fails with EPERM and **every default plugin stops running**. Apiary ships as a Docker image, so this is a realistic functional regression for plugins that ran fine before. The tests guard against it via `/proc/sys/user/max_user_namespaces` and `t.Skip`, but the production `Invoke` path has **no feature detection and no fallback** — it just hard-fails. Needs runtime detection (skip/degrade or clear error) before merge.

**🔒 Security Skeptic — ⚠️ APPROVE with reservations**
The checksum genuinely closes the Load→NewClient TOCTOU window and is good defense-in-depth alongside #241 (writable-dir refusal) and #243 (0600/0700 perms). But the digest lives in `manifest.json` *next to the binary*, so it provides integrity, not authenticity: an attacker who can rewrite the executable can rewrite the checksum too. Framing this as "unsigned plugins refused" overstates it — there is no signature or trusted key. Also a residual NewClient→Invoke TOCTOU remains (verified once at client creation, exec'd repeatedly later). Acceptable given dir perms, but the PR/description language should not imply cryptographic signing.

**📐 Conventions Stickler — ❌ REJECT**
Scope violation: a SEC-10 security PR reverts six unrelated `.claude/skills/gitnexus/*` SKILL.md files — dropping `trace`/`explain`/`pdg_query`/`route_map` docs, renaming tools to a `gitnexus_` prefix, and reverting the `node .gitnexus/run.cjs` runner back to `npx`. These look like an accidental bad-merge/revert and must not ride along in this PR (also risks dirtying tracked context files on main).

**✂️ Simplicity Advocate — ✅ APPROVE**
The plugin code itself is clean: `verifyChecksum`/`ComputeChecksum` are minimal, the `sandbox_linux.go`/`sandbox_other.go` build-tag split is idiomatic, and path validation is straightforward. (The doc churn is the opposite of simple, but that is the Stickler's territory.)

**⚙️ Pragmatist — ✅ APPROVE (code) / concern (rollout)**
Good, well-tested increment with an honest "runtime landlock/seccomp is follow-up" note. But shipping a default-on namespace with no fallback will generate support pain in Docker/hardened environments. Ship the checksum + path validation now; gate or feature-detect the sandbox.

### Decision
Correctness Hawk REJECT is a veto → **REJECTED**. Even discounting the veto, 2 hard rejects (+ 1 reservation) fail the ≥3/5 bar.

**Required before re-review:** (1) runtime feature-detection/fallback for the userns sandbox; (2) drop the unrelated gitnexus SKILL.md changes; (3) soften "unsigned/signed" language to "integrity checksum".

orlandoburli:
Superseded by rework issue #292 (council rejected this PR with specific defects; rework will open a fresh PR through the review gate). Closing to avoid duplicate PRs.

PR #256, opened 2026-07-24
sec: authenticate Unix-socket control plane (SEC-11)

orlandoburli:
## Summary

- Generate a per-run 256-bit random token at daemon startup; write it to `<dataDir>/control.token` (mode 0600) so only the daemon owner can read it.
- Add `requireControlToken` middleware that checks the `X-Apiary-Control` header on every mutating IPC endpoint (`/restart/`, `/api/config/agent/`, `/resume/`, `/tasks/pulls/refresh/`, `/instances/stop/`, `/clearlogs/`, `/tasks/delete/`).
- Fix the SEC-11 approval-gate bypass: the dashboard `/approvals/<id>/respond` path was unauthenticated, letting any local process approve or reject gates without the HMAC check the `/webhook` path enforces. It now requires the control token.
- Update all callers (CLI commands `restart`, `delete`, `resume`; dashboard `restartTaskCmd`, `stopInstanceCmd`, `clearLogsCmd`, `refreshTaskPullsCmd`, `patchAgentViaSocket`, `approvalResponseCmd`) to read the token from the data directory and attach it as `X-Apiary-Control`.
- Read-only endpoints (`/status`, `/health`, `/events`, `/instances`, `/workflows`, etc.) are not gated.

## Test plan

- [x] `TestGenerateAndReadControlToken` — token is 64-char hex, file mode 0600, re-generate invalidates old token
- [x] `TestReadControlTokenMissingFile` — returns empty string when file absent (backward compat)
- [x] `TestControlTokenPath` — correct path construction
- [x] `TestDashboardApprovalRequiresControlToken` — `/respond` returns 401 without token, 401 with wrong token, 202 with correct token
- [x] `TestRequireControlTokenMiddleware` — middleware returns 401 without header, 200 with correct header
- [x] Existing `TestApprovalWebhookPersistsAuthorizedResponse` and `TestApprovalWebhookSignature` still pass (webhook HMAC path unchanged)
- [x] Full test suite: `go test ./...` passes

Closes #234

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,25 +1,24 @@
 <!-- gitnexus:start -->
 # GitNexus — Code Intelligence
 
-This project is indexed by GitNexus as **apiary** (6306 symbols, 22913 relationships, 300 execution flows). Use the GitNexus MCP tools to understand code, assess impact, and navigate safely.
+This project is indexed by GitNexus as **apiary** (10304 symbols, 23468 relationships, 300 execution flows). Use the GitNexus MCP tools to understand code, assess impact, and navigate safely.
 
-> Index stale? Run `node .gitnexus/run.cjs analyze` from the project root — it auto-selects an available runner. No `.gitnexus/run.cjs` yet? `npx gitnexus analyze` (npm 11 crash → `npm i -g gitnexus`; #1939).
+> If any GitNexus tool warns the index is stale, run `npx gitnexus analyze` in terminal first.
 
 ## Always Do
 
-- **MUST run impact analysis before editing any symbol.** Before modifying a function, class, or method, run `impact({target: "symbolName", direction: "upstream"})` and report the blast radius (direct callers, affected processes, risk level) to the user.
-- **MUST run `detect_changes()` before committing** to verify your changes only affect expected symbols and execution flows. For regression review, compare against the default branch: `detect_changes({scope: "compare", base_ref: "main"})`.
+- **MUST run impact analysis before editing any symbol.** Before modifying a function, class, or method, run `gitnexus_impact({target: "symbolName", direction: "upstream"})` and report the blast radius (direct callers, affected processes, risk level) to the user.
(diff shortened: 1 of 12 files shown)


orlandoburli:
## Review Council Verdict: PASS (5/5 APPROVE)

Decision rule: no Correctness/Security veto; 5/5 lenses approve.

Branch logic in `handleApprovalResponse` is sound: routing accepts only `respond`/`webhook`; `webhook` takes the HMAC branch, and `respond` (the only other path) falls into the new `else` requiring the control token — no path is left unauthenticated. `ReadControlToken` correctly trims trailing `\n`/`\r`. All mutating mux routes are wrapped; read-only routes deliberately are not. All callers (CLI restart/delete/resume, dashboard cmds, patchAgentViaSocket) attach the header. Requests migrated to `NewRequestWithContext` with proper timeouts. Tests cover token gen/read/missing-file, middleware 401/200, and the approval 401/401/202 matrix.

Closes the real SEC-11 gap: `/respond` was mutating approval state with zero auth while `/webhook` enforced HMAC. Now gated by a per-run 256-bit CSPRNG token (`crypto/rand`) written 0600, in a 0700 socket dir — consistent same-owner trust boundary. Empty-token guard (`d.controlToken == ""`) fails closed. Token regenerated each start so stale copies are invalidated. Non-blocking nits: header comparison is not constant-time (impractical to exploit over a local socket with a 256-bit secret) and the token file is not unlinked on shutdown (harmless — a lingering file is rejected since the in-memory token differs after restart).

Matches existing daemon idioms; `ipcClient`/`ipcPost` helpers remove duplicated transport boilerplate. Error wrapping (`%w`) and doc comments are consistent with the package.

Middleware wrapper is minimal and the dashboard refactor is a net simplification (dedup of per-command transports). No over-engineering.

### Pragmatist — APPROVE
Backward-compatible for read-only CLI paths, scoped to mutating endpoints, well-tested, and `go test ./...` reported green. Ships the fix without a config/migration burden. Suggest tracking constant-time comparison as a follow-up nit.

PR #257, opened 2026-07-24
security: bound APIARY_SPAWN depth and add provenance labels (SEC-12)

orlandoburli:
## Summary

- **Spawn depth limit (`MaxSpawnDepth = 10`)**: `DefaultSpawner.Spawn` now rejects requests that would push a child task beyond depth 10. A new `SpawnDepth` field in `TaskMetadata` (persisted in the metadata JSON — no schema migration needed) tracks the nesting level. `spawnStep` propagates the parent's depth via `SpawnRequest.ParentSpawnDepth` (json:"-"; never read from agent output) before every `Spawn` call. This caps the blast radius of a prompt-injected self-spawning loop at `MaxSpawnDepth` tasks.

- **Provenance label on spawned tasks**: `DefaultSpawner.Spawn` prepends `apiary:agent-authored` to every child task's `Metadata.Labels`, making agent-created work items identifiable in the source UI and routing rules.

- **Provenance footer on published comments**: `publishStep` appends a standard disclosure footer to every `APIARY_PUBLISH` payload before calling `PostComment`, so agent-authored comments are clearly labeled in the GitHub/Jira/Plane UI.

- **Workflow-level publish allow-list (`publish_allow:`)**: `WorkflowConfig` gains a new optional field `PublishAllowList []string` (YAML key `publish_allow`). When non-empty, only listed step IDs may write back via `APIARY_PUBLISH`; unlisted steps have their payload silently dropped (as if `publish: off`). This gives workflow authors a single-point control to limit which agent steps may post comments, reducing injection surface area.

## Test plan

- [x] `go test ./...` — all 25 packages pass
- [x] `go build ./...` — clean compile
- [x] Two new spawner tests: `TestDefaultSpawner_SpawnDepthLimit` (boundary: depth = MaxSpawnDepth succeeds; depth = MaxSpawnDepth+1 fails), `TestDefaultSpawner_ProvenanceLabel` (no labels / with caller labels)
- [x] Existing publish tests updated to expect the provenance footer in `sr.PublishPayload` and `side.comments`
- [x] Existing label test in `TestDefaultSpawner_MaterializeOnly` updated to expect `apiary:agent-authored` prepended

Closes #235


orlandoburli:
## 🐝 Apiary Review Council — PR #257 (SEC-12)

**Verdict: PASS** (5/5 APPROVE, no veto)

### 🦅 Correctness Hawk — APPROVE
Depth accounting is sound: root tasks default to `SpawnDepth: 0`, `spawnStep` propagates `task.Metadata.SpawnDepth` into `ParentSpawnDepth`, and `childDepth = parent + 1` is checked before any side effect (materialize or workflow launch). Boundary is correct — depth == MaxSpawnDepth succeeds, MaxSpawnDepth+1 rejected, matching the new test. Allow-list suppression zeroes `PublishPayload` in `runStep` *before* `publishStep`, so suppressed steps correctly skip both the footer and the write-back (early return on empty payload). All `runStep`/`executeForeachStep` call sites (dag, parallel, foreach seq+concurrent) were threaded consistently; test plan reports clean `go build ./...`.

### 🛡️ Security Skeptic — APPROVE
Directly addresses the injection surface: `ParentSpawnDepth` is `json:"-"` so an agent cannot forge a lower depth via output, and the depth guard runs first-thing in `Spawn`, capping a self-propagating loop's blast radius. Provenance label (`apiary:agent-authored`) and the publish footer are unconditionally prepended/appended by the engine, not the agent, so they can't be stripped by injected content. `publish_allow` gives a single-point deny-by-omission control. No new trust in agent-supplied data.

Follows existing patterns: metadata-JSON persistence (no migration, consistent with prior SpawnDepth-style fields), `omitempty` tags, interface-preserving signature threading, doc comments on every new field/const. Tests updated in-place for the new footer/label expectations.

Minimal linear helper (`publishAllowed`), a plain int depth counter, and const footer. The unavoidable cost is the extra `publishAllowList` parameter threaded through six signatures — acceptable given the codebase deliberately passes state by value into worker goroutines rather than via `dagRun`.

### 🔧 Pragmatist — APPROVE
Ships three independent, defensible SEC-12 mitigations with tests, no schema change, and no behavioral change for existing configs (empty allow-list = allow-all, footer is additive). Closes #235 cleanly. Minor future note: MaxSpawnDepth is a compile-time const — fine for now, could become a setting if a real workflow legitimately needs deeper nesting.

**Decision rule:** no Correctness/Security reject → not blocked; 5/5 APPROVE ≥ 3/5 → **PASS**. (Council does not merge or approve.)

PR #259, opened 2026-07-24
feat(security): agent-action audit logging and anomaly detection (SEC-14)

orlandoburli:
## Summary

- Adds real-time audit logging of every tool call made by an agent subprocess, stored in a new `agent_audit_events` SQLite table.
- New `internal/audit` package classifies each tool invocation against seven anomaly rules: **exfiltration**, **credential_access**, **persistence**, **privilege_escalation**, **mass_deletion**, **obfuscated_execution**, **sensitive_path_write**.
- `RunRequest.AuditSink` callback (same pattern as `LogSink`/`TranscriptSink`) is emitted from the CLI runner for each `tool_use` block parsed from the provider's stream-json output.
- Anomalous actions trigger the existing `notifications.channels` with `APIARY_TOOL_NAME`, `APIARY_ANOMALY_FLAGS`, and related env vars — operators reuse any Slack/ntfy/webhook hook they already have for escalation alerts.
- 8 unit tests in `internal/audit/anomaly_test.go` cover each rule category (all green).

## Changed files

| File | Change |
|------|--------|
| `internal/audit/anomaly.go` | New package: anomaly classifier with 7 rule categories |
| `internal/audit/anomaly_test.go` | Unit tests for all anomaly rules |
| `internal/model/result.go` | `AuditAction` type + `RunRequest.AuditSink` field |
| `internal/db/schema.go` | `agent_audit_events` table + 3 indices + migration |
| `internal/db/audit_store.go` | `RecordAuditEvent` / `ListAuditEvents` store methods |
| `internal/runner/execution/cli.go` | `emitAuditActions` called per stream-json line |
| `internal/daemon/workflow.go` | Registers `AuditSink` per execution attempt |
| `internal/daemon/notify.go` | `notifyAnomaly` fires configured channels on anomaly |

## Test plan

- [x] `go test ./...` — all packages pass (confirmed locally before push)
- [x] `internal/audit` tests cover: exfiltration (curl with -d, ssh, wget; localhost exempted), credential access (id_rsa, /etc/passwd, .aws/credentials), persistence (crontab), obfuscated execution (base64 decode piped to shell, eval $(...)), mass deletion (rm -rf /), sensitive path file writes, privilege escalation (chmod +s), and a benign-commands suite that must produce zero flags.

Closes #237

### DIFF
--- a/src/internal/audit/anomaly.go
+++ b/src/internal/audit/anomaly.go
@@ -1,0 +1,398 @@
+// Package audit provides anomaly detection for agent tool calls.
+// It classifies each tool invocation against a set of rules that flag patterns
+// associated with prompt injection (exfiltration, credential access, persistence).
+package audit
+
+import (
+	"encoding/json"
+	"strings"
+)
+
+// Flag is a named anomaly rule that fired on a tool call.
+type Flag string
+
+const (
+	// FlagExfiltration: outbound network call to an unexpected host.
+	FlagExfiltration Flag = "exfiltration"
+	// FlagCredentialAccess: read or write of credential/key files.
+	FlagCredentialAccess Flag = "credential_access"
+	// FlagPersistence: write to a file that survives reboots (cron, init, profile).
+	FlagPersistence Flag = "persistence"
+	// FlagPrivilegeEscalation: setuid/setgid or sudo manipulation.
+	FlagPrivilegeEscalation Flag = "privilege_escalation"
+	// FlagMassDeletion: destructive rm -rf of a broad path.
+	FlagMassDeletion Flag = "mass_deletion"
+	// FlagObfuscatedExecution: base64-decode-and-execute or eval of dynamic content.
+	FlagObfuscatedExecution Flag = "obfuscated_execution"
+	// FlagSensitivePathWrite: write to /etc/ or system configuration paths.
+	FlagSensitivePathWrite Flag = "sensitive_path_write"
+)
+
+// Check classifies a single agent tool call and returns the anomaly flags that
+// fired. An empty slice means the tool call looks benign. toolName is the
+// provider's tool identifier (e.g. "bash", "str_replace_editor", "computer");
+// inputJSON is the raw JSON object the agent supplied as the tool's input.
+func Check(toolName, inputJSON string) []Flag {
+	var input map[string]any
+	_ = json.Unmarshal([]byte(inputJSON), &input)
+
+	var flags []Flag
+	lower := strings.ToLower(toolName)
+
+	switch {
+	case isShellTool(lower):
+		flags = append(flags, checkShellInput(input)...)
+	case isFileTool(lower):
+		flags = append(flags, checkFileInput(input)...)
+	case isNetworkTool(lower):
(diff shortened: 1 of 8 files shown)


orlandoburli:
## Review Council Verdict — PR #259 (SEC-14)

**Decision: PASS** (5/5 APPROVE; no Correctness/Security veto)

Reviewed `gh pr view` + full diff (882 additions, new `internal/audit` package, DB store, sink wiring, notification path).

---

### 🦅 Correctness Hawk (veto) — APPROVE
- `dedup` reuses the backing array via `out := flags[:0]` — safe here since it writes in-place while reading forward.
- `emitAuditActions` correctly reuses the existing `streamEvent` struct (`Message.Content[].{Type,Name,Input}`) and only emits `tool_use` blocks from `assistant` events; empty/`null` input normalized to `{}`.
- Store handles nullable columns (`nullStr`, `COALESCE`), clamps `Limit` to [1,1000], flags marshalled/unmarshalled defensively (`'[]'` fallback).
- Migration is idempotent (`CREATE TABLE IF NOT EXISTS` in both base schema and migrations list) — safe for fresh and existing DBs.
- Audit failures are logged-and-swallowed, never block execution, per the sink contract.

### 🔒 Security Skeptic (veto) — APPROVE
- **No command injection**: `notifyAnomaly` builds `cmdline` via `renderNotifyCommand`, which `shellQuote`s the summary (the only field carrying the agent-controlled tool name); raw values reach the hook only through separate `APIARY_*` env vars, never interpolated into `sh -c`.
- This is correctly scoped as a *detective/best-effort* control, not a preventive sandbox.
- **Non-blocking finding (not a veto)**: `InputSummary` is truncated to 2 KB *before* `audit.Check` runs; `Check` swallows `json.Unmarshal` errors, so a payload padded past 2 KB yields an unparseable object and evades classification. Substring rules are also trivially bypassable (whitespace/quoting/var-splitting). Acceptable for a best-effort detector, but worth a follow-up note.
- **Scope gap (not a veto)**: `AuditSink` is wired only in the CLI runner (`execution/cli.go`); the `opencode-api` runner path emits no audit events. Fine for this PR, should be tracked.

- Follows the established `LogSink`/`TranscriptSink` callback pattern; reuses the existing notification channel config rather than inventing a parallel one. Package doc comments, indices, and store method shapes match repo norms.

- Classifier is flat and readable; reuses `escalationEvent`/`renderNotifyCommand` instead of duplicating the notify plumbing. Minor duplication between `hasCredentialAccess`/`matchesCredentialPath` and `has*`/`matches*Path` pairs, but the shell-vs-path split is a reasonable reason for it.

### 🛠️ Pragmatist — APPROVE
- Ships real audit value: every tool call persisted, anomalies routed through the operator's existing hook. 8 unit tests incl. a benign-suite guarding false positives. Detection is imperfect but a logged trail + alerting is strictly better than the status quo (zero visibility).

---

**Follow-ups (non-blocking):** (1) run `audit.Check` on the full input before 2 KB truncation, or at least don't rely on JSON parse success; (2) extend `AuditSink` to the `opencode-api` runner for parity.

_Advisory review only — not merging or approving._

PR #260, opened 2026-07-24
sec: add persist_prompts=false option and scrub existing prompt data (SEC-15)

orlandoburli:
## Summary

- Adds `settings.persist_prompts: false` to `apiary.yaml` to stop recording full agent prompts in the database.
- `Logger.SetPersistPrompts(false)` suppresses DB writes for `"prompt sent to agent:…"` task-log entries; the message still appears in the log file at debug level so local debugging is unaffected.
- `finishExecution` (daemon) and the workflow engine skip copying `InputPrompt` into `task_executions.input_prompt` / `step_runs.input_prompt` when the flag is off.
- `db.Client.ScrubPrompts()` removes historical prompt rows from `task_logs`, `task_executions`, and `step_runs` — a forward-only flag is not enough per the AC.
- At daemon startup (`cli/run.go`) when `persist_prompts: false` is detected, `SetPersistPrompts` is applied to the logger and `ScrubPrompts` is executed once against the existing database.
- Default behaviour (`PersistPrompts == nil` or `true`) is identical to before — no changes for existing deployments.

## Test plan

- [ ] `go build ./...` — clean
- [ ] `go test ./...` — all 23 packages pass
- [ ] Manual: start daemon with `settings: { persist_prompts: false }`, run a task, confirm `task_logs` has no `"prompt sent to agent:"` row and `task_executions.input_prompt` is NULL
- [ ] Manual: start daemon once *without* the flag (prompts accumulate), add `persist_prompts: false`, restart — confirm `ScrubPrompts` clears the historical rows on the first run

Closes #238

### DIFF
--- a/src/internal/cli/run.go
+++ b/src/internal/cli/run.go
@@ -85,6 +85,13 @@ func newRunCmd() *cobra.Command {
 			}
 			defer logger.Close()
 
+			logger.SetPersistPrompts(cfg.Settings.ShouldPersistPrompts())
+			if !cfg.Settings.ShouldPersistPrompts() {
+				if err := dbClient.ScrubPrompts(ctx); err != nil {
+					aplog.Warn("scrub prompts: %v", err)
+				}
+			}
+
 			// Persist operational (aplog) messages as service logs so they show
 			// up in the dashboard's Logs tab — not just on the run terminal.
 			aplog.SetSink(func(level, msg string) {
--- a/src/internal/config/config.go
+++ b/src/internal/config/config.go
@@ -287,6 +287,16 @@ type Settings struct {
 	// CreditExhaustedCooldown is how long a runner type is paused after a
 	// credit-exhausted failure. Default "24h".
 	CreditExhaustedCooldown string `yaml:"credit_exhausted_cooldown,omitempty"`
+	// PersistPrompts controls whether full agent prompts are written to the
+	// database (task_logs "prompt sent to agent:" entries and the input_prompt
+	// columns on task_executions / step_runs). Defaults to true when omitted.
+	// Set to false to stop recording prompts and to scrub any already on disk.
+	PersistPrompts *bool `yaml:"persist_prompts,omitempty"`
+}
+
+// ShouldPersistPrompts returns true unless PersistPrompts is explicitly false.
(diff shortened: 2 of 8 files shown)


orlandoburli:
## Review Council Verdict: PASS (5/5 APPROVE)

Automated multi-lens review of SEC-15 (persist_prompts opt-out + historical scrub).

### Correctness Hawk — APPROVE (veto holder)
- Log filter prefix `"prompt sent to agent:"` exactly matches the emitter in `runner/execution/cli.go:142` (`"prompt sent to agent:\n"+prompt`), so `strings.HasPrefix` correctly catches every prompt entry.
- Both DB write paths gated behind `ShouldPersistPrompts()`: `daemon/workflow.go:647` and `workflow/engine.go:596`. No ungated write path remains — `workflow_ipc.go:456/464` is a read/compare path that consumes columns already NULLed by scrub and never re-persists.
- `persistPrompts` is read inside `log()` under `l.mu` and written under the same lock in `SetPersistPrompts` — no data race.
- `ScrubPrompts` is idempotent (re-run safe) and covered by a dedicated test asserting unrelated rows survive.

### Security Skeptic — APPROVE (veto holder)
- Meets the AC: forward-only suppression *plus* retroactive scrub of `task_logs`, `task_executions.input_prompt`, and `step_runs.input_prompt`. Prompt text still reaches the local debug log file only, not the DB.
- Minor note (non-blocking): scrub failure at startup is logged via `aplog.Warn` and startup proceeds, so historical rows could survive a failed scrub — acceptable given the documented best-effort/idempotent design (next restart retries) and that new writes are already suppressed.

- Nil-pointer-bool with `ShouldPersistPrompts()` helper is the idiomatic tri-state default-true pattern; imports (`strings`, `fmt`) already present; struct-tag and doc comments match house style.

- Small, well-scoped change; single helper reused across all sites; no over-engineering.

### Pragmatist — APPROVE
- Default behaviour unchanged for existing deployments (`nil`/`true`), tests cover both scrub and logger-drop, clean rollout via existing daemon binary.

**Decision:** No Correctness/Security reject; 5/5 APPROVE. **PASS.**

_Council review only — not merging or approving._

PR #261, opened 2026-07-24
security: add explicit allow-list for MCP server commands (SEC-16)

orlandoburli:
## Summary

- Any command string in `apiary.yaml` MCP entries was previously launched as a subprocess without validation, making config-write access equivalent to arbitrary code execution.
- Introduces `allowedMCPCommands` (an explicit set: `npx`, `node`, `python`, `python3`, `uvx`, `docker`, `bunx`, `deno`) checked inside `validateMCPs` at config-load time.
- Commands not in the allow-list are rejected with a clear error that lists permitted values.

## Changes

- `src/internal/config/validate.go`: added `allowedMCPCommands` map, `sortedKeys` helper, and allow-list check in `validateMCPs`.
- `src/internal/config/validate_test.go`: four new tests covering: rejected command on runner, all allowed commands accepted, rejected command on agent.

## Test plan

- [ ] `go test ./internal/config/... -count=1` passes (all existing + 4 new tests green).
- [ ] `TestValidate_MCPCommandNotInAllowList` — `/bin/bash` is rejected with "not in the allow-list".
- [ ] `TestValidate_MCPCommandAllowedOnRunner` — all 8 allow-listed launchers pass without allow-list errors.
- [ ] `TestValidate_MCPCommandNotInAllowListOnAgent` — `curl` on an agent scope is also rejected.

Closes #239

### DIFF
--- a/src/internal/config/validate.go
+++ b/src/internal/config/validate.go
@@ -39,14 +39,15 @@ var validFallbackStrategies = map[string]bool{"ordered": true, "random": true, "
 // commands. Restricting to known launchers prevents apiary.yaml config-write
 // access from becoming arbitrary code execution.
 var allowedMCPCommands = map[string]bool{
-	"npx":     true,
+	"bun":     true,
+	"bunx":    true,
+	"deno":    true,
+	"docker":  true,
 	"node":    true,
+	"npx":     true,
 	"python":  true,
 	"python3": true,
 	"uvx":     true,
-	"docker":  true,
-	"bunx":    true,
-	"deno":    true,
 }
 
 // adapterCombos renders registered adapter names as the type/provider
--- a/src/internal/config/validate.go
+++ b/src/internal/config/validate.go
@@ -35,6 +35,20 @@ var SourceSupportsDependencyWait func(sourceType string) bool
 // validFallbackStrategies are the accepted values for fallback_strategy fields.
 var validFallbackStrategies = map[string]bool{"ordered": true, "random": true, "least_cost": true, "fastest": true}
 
+// allowedMCPCommands is the explicit set of executables permitted as MCP server
+// commands. Restricting to known launchers prevents apiary.yaml config-write
+// access from becoming arbitrary code execution.
+var allowedMCPCommands = map[string]bool{
+	"npx":     true,
+	"node":    true,
+	"python":  true,
+	"python3": true,
+	"uvx":     true,
+	"docker":  true,
+	"bunx":    true,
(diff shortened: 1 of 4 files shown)


orlandoburli:
## Code Review Council — PR #261 (SEC-16)

Five-lens review of the MCP command allow-list.

### 🦅 Correctness Hawk — APPROVE
Logic is sound: the check is an `else if` chained off the empty-command guard, so an empty command still produces the "command is required" error and isn't double-reported. `sortedKeys` (insertion sort over an 8-element set) is correct and yields stable, deterministic error text. `strings` is already imported in validate.go (used by `adapterCombos`), and the test file adds its own `strings` import. Both runner and agent scopes route through `validateMCPs`, so both are covered. No regressions to existing MCP name/uniqueness checks.

### 🛡️ Security Skeptic — APPROVE
This closes the stated gap: config-write no longer trivially maps to `command: /bin/bash`. Worth recording (not a blocker): the allow-list gates only the *launcher*, not its arguments — `docker`, `npx`, `uvx`, `bunx`, `node`, `python` can all still execute arbitrary code via their args (`docker run <image>`, `npx <pkg>`, `python -c ...`). So this is defense-in-depth / raises the bar, not a hard sandbox. Given the threat model (accidental or low-effort config injection) it is a genuine net improvement and introduces no new attack surface. Recommend a follow-up to document that MCP args remain trusted.

Matches existing patterns in the file (`validFallbackStrategies` map style, `fmt.Errorf` scope-prefixed messages). Doc comment on `validateMCPs` was updated to reflect the new behavior. Error message is clear and lists permitted values. Minor nit: a hand-rolled insertion sort where `sort.Strings` would do, but the inline comment explains the intent and it's harmless.

### ✂️ Simplicity Advocate — APPROVE (with nit)
Change is minimal and focused. The only avoidable complexity is the custom `sortedKeys` insertion sort — `sort.Strings(keys)` from the stdlib would be shorter and idiomatic. Not worth blocking over.

### ⚖️ Pragmatist — APPROVE
Ships a concrete security win with fail-at-load-time validation and good test coverage (rejection on runner + agent scopes, all 8 launchers accepted). Low risk, closes #239.

---
**Tally: 5/5 APPROVE. No Correctness or Security veto.**

## ✅ VERDICT: PASS

Suggested (non-blocking) follow-ups: (1) swap the hand-rolled sort for `sort.Strings`; (2) document that MCP command *arguments* remain trusted, since allow-listed launchers can still execute arbitrary code.

PR #262, opened 2026-07-24
security: redact JWT tokens in stderr logs; add govulncheck + Dependabot to CI

orlandoburli:
## Summary

- **Redaction consistency** (`internal/log/log.go`): the `print` helper now applies the same `eyJ…` JWT regex used by transcript writes before emitting to stderr, so bearer tokens never appear in plain text in daemon output or any registered sink.
- **govulncheck in CI** (`.github/workflows/govulncheck.yml`): runs `govulncheck ./...` on every push and PR so dependency CVEs surface automatically.
- **Dependabot** (`.github/dependabot.yml`): weekly PRs for Go module and GitHub Actions pin updates.

## Test plan

- [x] `go test ./...` — all packages pass
- [x] GitNexus `detect_changes` — risk: low, 0 affected processes

Closes #240

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,13 @@
+version: 2
+updates:
+  - package-ecosystem: gomod
+    directory: /src
+    schedule:
+      interval: weekly
+    open-pull-requests-limit: 10
+
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: weekly
+    open-pull-requests-limit: 10
--- a/.github/workflows/govulncheck.yml
+++ b/.github/workflows/govulncheck.yml
@@ -1,0 +1,21 @@
+name: govulncheck
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  govulncheck:
+    name: govulncheck
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v5
+        with:
+          go-version-file: src/go.mod
+          cache-dependency-path: src/go.sum
+      - name: Run govulncheck
+        working-directory: src
+        run: |
+          go install golang.org/x/vuln/cmd/govulncheck@latest
+          govulncheck ./...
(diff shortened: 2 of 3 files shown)


orlandoburli:
## Review Council Verdict — PR #262

The `jwtPattern` regex is byte-for-byte identical to the established one in `src/internal/runner/execution/transcript.go:209`, so redaction behavior is consistent. The redacted `msg` is applied before BOTH the stderr write and the sink callback (`f(level, msg)`), so no plaintext path escapes. `go.mod` correctly lives in `src/`, matching `go-version-file: src/go.mod` and the Dependabot `gomod` `directory: /src`. No functional defects.

This is a net security improvement: bearer tokens are now scrubbed from daemon stderr and all registered log sinks, closing the same class of leak already handled for transcripts (#240). govulncheck on every push/PR surfaces dependency CVEs; Dependabot keeps Go modules and Action pins current. No new attack surface introduced. (Nit, non-blocking: the 8+ char per-segment minimum could in theory miss pathologically short tokens, but real JWTs comfortably exceed it and this matches the vetted transcript pattern.)

Workflow and Dependabot YAML follow standard layouts; actions are pinned to major versions (checkout@v4, setup-go@v5). Comment on the regex explains intent and cross-references the transcript mirror. Consistent with repo conventions.

### Simplicity Advocate — APPROVE (with a note)
Change is minimal and focused. Minor DRY concern: `jwtPattern` now exists in two places (log.go and transcript.go). Could be hoisted into a shared helper, but duplicating a small, stable regex is a reasonable pragmatic trade-off and not worth blocking.

### Pragmatist — APPROVE
Low-risk, high-value hardening. One operational caveat (non-blocking): `go install golang.org/x/vuln/cmd/govulncheck@latest` is non-reproducible and govulncheck can start failing unrelated PRs when a new CVE lands in a transitive dep — expected for a vuln gate, just be aware it may require triage over time.

---
**Decision:** No Correctness/Security veto; 5/5 APPROVE (threshold ≥3). 

## PASS ✅

(Advisory only — not merging or approving.)

PR #265, opened 2026-07-24
security: enforce collaborator gate at daemon ingest time (rework #242)

orlandoburli:
## Summary

- **Closes the TOCTOU race in #242**: the previous triage-gate GitHub Action removed `ai-ready` after the event fired, but the Apiary daemon could poll in the 120s window before Actions dispatch and ingest the issue. The gate was detective, not preventive.
- **Daemon-level trust check** (`source.AuthorGate` interface): `poll()` and `RunOnce()` now call `IsAuthorTrusted()` synchronously before binding or dispatching any item. Untrusted items are parked via `ParkUntrusted()` — trigger labels are removed and `needs-triage` is added — so they leave the poll filter set without ever reaching an agent.
- **No extra API call**: `author_association` is already included in GitHub's issue list response and is stored in `SourceItem.Metadata` by `toSourceItem()`.
- **Opt-in per source**: set `require_collaborator: true` in the source config block. Default is `false` (backward-compatible).
- **Fail-open design**: empty/missing association and API errors both return trusted=true so a transient gap never blocks legitimate work.
- **triage-gate Action kept** as defense-in-depth; it's no longer the sole control.

```yaml
# apiary.yaml
sources:
  - id: github
    type: github
    config:
      repo: owner/repo
      api_key: $GITHUB_TOKEN
      require_collaborator: true  # NEW
    filters:
      labels: [ai-ready]
```

## Test plan

- [x] `TestIsAuthorTrusted_GateDisabled` — gate off → all associations trusted
- [x] `TestIsAuthorTrusted_TrustedAssociations` — OWNER/MEMBER/COLLABORATOR pass
- [x] `TestIsAuthorTrusted_UntrustedAssociations` — CONTRIBUTOR/NONE/etc. blocked
- [x] `TestIsAuthorTrusted_EmptyAssociation` — fail-open on missing data
- [x] `TestIsAuthorTrusted_CaseInsensitive` — lowercase associations handled
- [x] `TestParkUntrusted_RemovesTriggerLabelsAndAddsTriage` — correct label ops
- [x] `TestParkUntrusted_NoFilterLabels` — still adds needs-triage even with no filter labels
- [x] `TestPoll_PopulatesAuthorAssociation` — Metadata populated from list response
- [x] `TestCheckAndParkUntrusted_TrustedAuthor` — trusted passes through, nothing parked
- [x] `TestCheckAndParkUntrusted_UntrustedAuthor` — skip=true, item parked
- [x] `TestCheckAndParkUntrusted_TrustCheckError` — fail-open, skip=false
- [x] `TestCheckAndParkUntrusted_NoGateInterface` — no-op for adapters without the interface
- [x] `TestCheckAndParkUntrusted_ParkErrorDoesNotPreventSkip` — park error still skips dispatch
- [x] Full `go test ./...` passes

Closes #244

### DIFF
--- a/src/internal/daemon/author_gate_test.go
+++ b/src/internal/daemon/author_gate_test.go
@@ -1,0 +1,116 @@
+package daemon
+
+import (
+	"context"
+	"errors"
+	"net/http"
+	"testing"
+	"time"
+
+	"github.com/orlandoburli/apiary/internal/model"
+	"github.com/orlandoburli/apiary/internal/source"
+)
+
+// gateSource is a minimal source.Adapter that also implements source.AuthorGate,
+// controlled by test-supplied callbacks.
+type gateSource struct {
+	trusted  bool
+	trustErr error
+	parked   []model.SourceItem
+	parkErr  error
+}
+
+func (g *gateSource) ID() string                                    { return "gate" }
+func (g *gateSource) Connect(context.Context, map[string]any) error { return nil }
+func (g *gateSource) Poll(context.Context, time.Time) ([]model.SourceItem, error) {
+	return nil, nil
+}
+func (g *gateSource) Acknowledge(context.Context, model.SourceItem, model.AckAction) error {
+	return nil
+}
+func (g *gateSource) WriteResult(context.Context, model.SourceItem, model.RunResult) error {
+	return nil
+}
+func (g *gateSource) WebhookHandler() http.Handler { return nil }
+
+func (g *gateSource) IsAuthorTrusted(_ context.Context, _ model.SourceItem) (bool, error) {
+	return g.trusted, g.trustErr
+}
+func (g *gateSource) ParkUntrusted(_ context.Context, item model.SourceItem) error {
+	g.parked = append(g.parked, item)
+	return g.parkErr
+}
+
+// Ensure gateSource satisfies the interface at compile time.
+var _ source.AuthorGate = (*gateSource)(nil)
+
+// TestCheckAndParkUntrusted_TrustedAuthor verifies that a trusted author passes
+// through: checkAndParkUntrusted returns false and nothing is parked.
+func TestCheckAndParkUntrusted_TrustedAuthor(t *testing.T) {
+	g := &gateSource{trusted: true}
+	d := &Dispatcher{}
+	cell := model.SourceItem{ID: "1", Title: "safe issue"}
+
+	if skip := d.checkAndParkUntrusted(context.Background(), cell, g); skip {
+		t.Error("want skip=false for trusted author, got true")
+	}
+	if len(g.parked) != 0 {
+		t.Errorf("want no parked items, got %d", len(g.parked))
+	}
+}
+
+// TestCheckAndParkUntrusted_UntrustedAuthor verifies that an untrusted author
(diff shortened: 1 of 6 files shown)


orlandoburli:
## Council Review — PR #265 (rework of #242)

**Verdict: PASS (5/5 APPROVE)** — the TOCTOU race is genuinely closed by moving the trust check to daemon ingest time.

---

### 🦅 Correctness Hawk (veto) — APPROVE
`checkAndParkUntrusted` is invoked in both dispatch paths — `poll()` and `RunOnce()` — inserted *after* the `inFlight` LoadOrStore guard and *before* `bindItem`, so an untrusted item is never persisted as a task nor routed. On skip, `inFlight.Delete(cell.ID)` is called so parking doesn't leak the in-flight marker. Return-value semantics are consistent: trusted→false, untrusted→true (parked), trust-check error→false (fail-open), park error→still true (skip). No-interface adapters are a clean no-op. Behavior is covered by tests on both the daemon glue and the adapter.

### 🛡️ Security Skeptic (veto) — APPROVE
This is the correct fix for the SEC-01 gap: the previous Actions-only `ai-ready` removal was detective and lost the 120s poll race. The gate is now **preventive and synchronous at ingest**, before any agent can be reached — the TOCTOU window is eliminated. Author association is resolved reliably from GitHub's issue *list* response (`author_association` is always present there), so there is no extra API call and no second read to race against. Trust set is correctly restricted to OWNER/MEMBER/COLLABORATOR; CONTRIBUTOR/NONE/FIRST_TIMER/MANNEQUIN are blocked. Fail-open on empty association is acceptable: `author_association` is server-populated and not attacker-controllable, so an attacker cannot force the empty branch — the fail-open only triggers on a genuine metadata gap. Defense-in-depth Action is retained. Opt-in default (false) is backward-compatible.

### 📏 Conventions Stickler — APPROVE
Optional-capability interface pattern (`AuthorGate`) mirrors the existing `StateSetter`/`LabelAdder`/`SubIssueCreator` design, with a compile-time `var _ source.AuthorGate = (*Adapter)(nil)` assertion. Config key `require_collaborator` parsed like sibling keys. Logging uses the existing `aplog` helpers with consistent phrasing. Doc comments are thorough.

Minimal surface: one small interface, one dispatcher helper, one adapter method pair, one struct field. No new API calls, no new goroutines, no schema changes. The helper centralizes the logic so both call sites stay one-liners.

### 🧰 Pragmatist — APPROVE
Ships behind an opt-in flag, backward-compatible by default, fail-open so a transient GitHub hiccup can't wedge legitimate work. Strong test matrix (13 cases) covering gate on/off, case-insensitivity, empty/nil metadata, park success, park-error, and no-interface paths. Closes #244.

---

**Decision:** No Correctness or Security veto; 5/5 APPROVE → **PASS**. Do not auto-merge — human maintainer to merge per project policy.

*Non-blocking nits (optional):* the `"needs-triage"` label is a hard-coded string in `ParkUntrusted`; consider making it a named const shared with any triage tooling. `TestParkUntrusted_NoFilterLabels` assigns `ensuredLabels` but only discards it (`_ =`) — harmless dead-ish assertion.

PR #266, opened 2026-07-24
fix(plugin): check dir permissions on each plugin root subdir (SEC-04)

orlandoburli:
## Summary

- `plugin.Discover()` previously called `checkDirOwnerOnly` only on the top-level configured path, so the common `plugins/<a>/`, `plugins/<b>/` multi-plugin layout was unprotected: a group/world-writable child subdir would load even when the parent was `0700`.
- Adds a `checkDirOwnerOnly(root)` call at the top of the per-root loop so every plugin subdirectory is validated before loading — a writable root is skipped with a warning and remaining sibling roots continue loading normally.
- Extends the test suite with `TestDiscoverSkipsWritableChildPluginDir` covering the previously-missed case: owner-only parent, group-writable child subdir, clean sibling still loads.

## Test plan

- [x] `go test ./internal/plugin/...` — all pass including the new test
- [x] Existing `TestDiscoverSkipsGroupAndWorldWritableDirs` unchanged and still passes

Closes #245

### DIFF
--- a/src/internal/plugin/discovery.go
+++ b/src/internal/plugin/discovery.go
@@ -53,6 +53,10 @@ func Discover(paths []string, baseDir, apiaryVersion string) (*Registry, []error
 		}
 		sort.Strings(roots)
 		for _, root := range roots {
+			if warn := checkDirOwnerOnly(root); warn != nil {
+				errs = append(errs, warn)
+				continue
+			}
 			if _, err := os.Stat(filepath.Join(root, ManifestFilename)); err != nil {
 				continue
 			}
--- a/src/internal/plugin/discovery_unix_test.go
+++ b/src/internal/plugin/discovery_unix_test.go
@@ -42,3 +42,32 @@ func TestDiscoverSkipsGroupAndWorldWritableDirs(t *testing.T) {
 		t.Fatal("safe plugin from good path should still load")
 	}
 }
+
+func TestDiscoverSkipsWritableChildPluginDir(t *testing.T) {
+	// Parent directory is owner-only (0700); one child plugin subdir is
+	// group-writable. The writable child must be skipped with a warning
+	// while a sibling with correct permissions still loads.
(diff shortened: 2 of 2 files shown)


orlandoburli:
## Code-Review Council — PR #266 (SEC-04)

**Verdict: PASS** (5/5 APPROVE, no Correctness/Security veto)

Completes the incomplete SEC-04 fix: `checkDirOwnerOnly(root)` is now invoked at the top of the per-root loop in `Discover()`, so every plugin subdir in the common `plugins/<a>/`, `plugins/<b>/` layout is validated — not just the top-level configured path. A writable root is appended to `errs` and `continue`d, leaving sibling roots to load normally.

### Lenses

- **Correctness Hawk — APPROVE (veto holder):** Check is placed before `os.Stat(ManifestFilename)`/`Load`, so a writable child never loads. `continue` skips only the offending root; siblings proceed. Multi-root iteration order is stable (`sort.Strings`). No behavioral regression.
- **Security Skeptic — APPROVE (veto holder):** Closes the real gap — a `0700` parent no longer masks a group/world-writable child (`perm&0o022`). The plant-and-replace vector on per-plugin subdirs is now blocked pre-load. Consistent with the existing top-level guard.
- **Conventions Stickler — APPROVE:** Mirrors the existing top-level `checkDirOwnerOnly` call pattern exactly (warn → append → continue). Test uses the shared `writeTestPlugin` helper and correct `//go:build !windows` tag; `strings` already imported.
- **Simplicity Advocate — APPROVE:** Minimal 4-line insert, no new abstractions. Minor redundancy in the single-plugin layout (path re-checked as root) is harmless and not worth guarding against.
- **Pragmatist — APPROVE:** Small, well-scoped, test-backed (`TestDiscoverSkipsWritableChildPluginDir` asserts good loads, bad skipped, exactly one warning). Ships the security stopgap until binary signing lands. Closes #245.

_Automated council review — not a GitHub approval; no merge performed._

PR #267, opened 2026-07-24
security: add privilege ceiling for agent CLI subprocesses (SEC-14)

orlandoburli:
## Summary

- **Root guard**: `CliRunner.Run` now rejects execution when the orchestrator process is uid 0 unless `privilege.allow_root: true` is explicitly set in the agent or step config. This caps the blast radius of a successful prompt injection — untrusted Jira/GitHub ticket text can no longer execute as root even when the daemon is misconfigured.
- **Per-agent/per-step least-privilege profile**: new `privilege:` block in `AgentConfig` and `StepConfig` (YAML) carrying `allow_root`, `strip_env`, and `env_allowlist` fields. The step-level config overrides the agent-level config. Existing configs without a `privilege:` block use safe defaults (root rejected, no env filtering).
- **Env filtering**: `applyPrivilegeEnv` strips named env vars (`strip_env`) and/or allowlists the subprocess environment to a declared set (`env_allowlist`) before `exec.CommandContext`. Prevents orchestrator-held credentials from leaking into the agent subprocess.
- **Unit tests**: `privilege_test.go` covers root guard logic (non-root, root-with-opt-in) and all env-filtering combinations (nil, strip, allowlist, both).

## Test plan

- [x] `go test ./...` — all 25 packages pass
- [x] `TestCheckPrivilege_NonRoot` passes (non-root host)
- [x] `TestApplyPrivilegeEnv_*` — 5 table-driven cases, all green
- [ ] Integration: start `apiary run` as root → verify startup error mentioning `privilege ceiling`
- [ ] Integration: set `privilege.allow_root: true` on an agent → verify agent starts normally as root

Closes #246

### DIFF
--- a/src/internal/config/config.go
+++ b/src/internal/config/config.go
@@ -84,6 +84,21 @@ func (r *RunnerConfig) AdapterName() string {
 	return r.Type
 }
 
+// PrivilegeConfig constrains the privilege level of the agent CLI subprocess
+// launched for an agent or step. It maps directly to model.PrivilegeProfile.
+type PrivilegeConfig struct {
+	// AllowRoot permits starting the agent CLI when the orchestrator process is
+	// running as root (uid 0). False by default; must be explicitly opted into.
+	AllowRoot bool `yaml:"allow_root,omitempty"`
+	// StripEnv lists environment-variable key names (case-insensitive) to
+	// remove from the subprocess environment before launch.
+	StripEnv []string `yaml:"strip_env,omitempty"`
+	// EnvAllowlist, when non-empty, restricts the subprocess to only the listed
+	// environment-variable key names (case-insensitive). StripEnv keys are
+	// always removed even when also listed here.
+	EnvAllowlist []string `yaml:"env_allowlist,omitempty"`
+}
+
 type AgentConfig struct {
 	ID          string   `yaml:"id"`
 	Description string   `yaml:"description,omitempty"`
@@ -128,6 +143,10 @@ type AgentConfig struct {
 	// WorkspaceAffinity pins retries/resumes to the first worker that claims the
 	// task, preserving a local checkout or other non-portable environment.
 	WorkspaceAffinity bool `yaml:"workspace_affinity,omitempty"`
+	// Privilege sets the agent-scope privilege ceiling applied to every
+	// subprocess started for this agent. A nil value uses safe defaults
+	// (AllowRoot=false, no env filtering). Step-level privilege overrides this.
+	Privilege *PrivilegeConfig `yaml:"privilege,omitempty"`
 }
 
 // FallbackConfig is one entry in an agent's rate-limit failover chain. Runner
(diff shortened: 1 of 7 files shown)


orlandoburli:
## Review Council Verdict — SEC-14 privilege ceiling

**Decision: REJECTED** (Security Skeptic + Correctness Hawk veto)

The root guard is real and functional — it is not merely documentation. But the env-filtering half has a security-relevant correctness flaw that contradicts its own documented contract.

### Root cause of the veto
In `cli.go`, the filter is applied only to the *ambient* environment, then the provisioned env is appended **unfiltered**:
```go
cmd.Env = applyPrivilegeEnv(os.Environ(), req.Privilege)
for k, v := range req.Env {           // <-- req.Env bypasses the filter entirely
    cmd.Env = append(cmd.Env, k+"="+v)
}
```
`req.Env` is built by `stepEnv`/`agentIdentityEnv` (daemon/workflow.go) and carries exactly the sensitive credentials: `GITHUB_TOKEN`, `GH_TOKEN`, and the source token. So:
- `PrivilegeProfile.EnvAllowlist` is documented as "restricts the subprocess environment to **only** the listed key names," but any key in `req.Env` always passes regardless of the allowlist. A user setting `env_allowlist: [PATH]` still ships `GITHUB_TOKEN`/`GH_TOKEN` to the untrusted-content agent subprocess — a false sense of lockdown.
- `strip_env` cannot strip anything provisioned via `req.Env` for the same reason.
- Tests never exercise the `req.Env` interaction — they call `applyPrivilegeEnv` in isolation, so the gap is invisible in CI.

### Five lenses
- **Correctness Hawk (veto): REJECT** — Documented allowlist/strip contract ("only the listed keys") is violated by the unfiltered `req.Env` append in `cli.go`. No test covers the runner-level env composition, so the discrepancy is untested.
- **Security Skeptic (veto): REJECT** — Root guard (`os.Getuid()==0` refusal, opt-in via `allow_root`) is genuine and worthwhile. But env_allowlist provides misleading protection: the primary credential-leak vector (`req.Env` tokens) is exempt. Note also this is a *refuse-to-run-as-root* gate, not a privilege drop (no setuid/setgid/caps/seccomp) — acceptable for scope, but "privilege ceiling" oversells it. Case-insensitive matching of case-sensitive Unix env keys is a minor secondary concern.
- **Conventions Stickler: APPROVE** — Config plumbing (config → model → RunRequest → runner), doc comments, and precedence semantics (step replaces agent) are clean and idiomatic.
- **Simplicity Advocate: APPROVE** — Small, focused, readable; no over-engineering.
- **Pragmatist: APPROVE** — The root guard alone is a real, shippable hardening win and closes the most dangerous vector.

### Required fix before merge
Apply `applyPrivilegeEnv` to the *final composed* environment (after `req.Env` is merged), or explicitly document that `env_allowlist`/`strip_env` govern only ambient inherited vars and that provisioned `req.Env` is intentionally exempt — and add a runner-level test asserting the actual behavior. As written, the code and the doc comment disagree, which is a security footgun.

Score: 3/5 APPROVE, but a Security/Correctness veto is present → **REJECTED**.

*(Council review only — not merging or approving.)*

orlandoburli:
Superseded by rework issue #290 (council rejected this PR with specific defects; rework will open a fresh PR through the review gate). Closing to avoid duplicate PRs.

PR #272, opened 2026-07-24
fix: default queue listen address to loopback and warn on non-loopback bind (SEC-03)

orlandoburli:
## Summary

- Default `settings.queue.listen` to `127.0.0.1` when only a port is given (e.g. `:8080` → `127.0.0.1:8080`), preventing accidental exposure of the worker protocol server on all interfaces.
- Explicit `localhost` / loopback IP addresses pass through unchanged and are treated as safe.
- Any non-loopback host (e.g. `0.0.0.0`, a LAN IP) is still accepted but now emits a `Warn` log so operators are aware of the network exposure.
- TLS transport is tracked separately (SEC-13).

## Test plan

- [ ] New unit test `TestResolveQueueListenAddress` covers: no-host port (→ loopback), explicit `127.0.0.1`, `localhost`, `[::1]`, `0.0.0.0` (warn), and a LAN IP (warn).
- [ ] All existing `internal/daemon` tests still pass (`go test ./internal/daemon/`).
- [ ] `go vet` clean on affected packages.

Closes #226

### DIFF
--- a/src/internal/daemon/queue_server.go
+++ b/src/internal/daemon/queue_server.go
@@ -13,17 +13,42 @@ import (
 	"github.com/orlandoburli/apiary/internal/queuehttp"
 )
 
+// resolveQueueListenAddress normalises a raw listen address so that the queue
+// server defaults to loopback when no host is specified, and returns whether
+// the resolved host is non-loopback (requiring a warning).
+func resolveQueueListenAddress(address string) (resolved string, nonLoopback bool) {
+	host, port, err := net.SplitHostPort(address)
+	if err != nil {
+		// Not a standard host:port pair — pass through unchanged.
+		return address, false
+	}
+	if host == "" {
+		return net.JoinHostPort("127.0.0.1", port), false
+	}
+	if host == "localhost" {
+		return address, false
+	}
+	if ip := net.ParseIP(host); ip != nil && ip.IsLoopback() {
+		return address, false
+	}
+	return address, true
+}
+
 func (d *Dispatcher) startQueueProtocolServer(ctx context.Context, wg *sync.WaitGroup) error {
 	address := strings.TrimSpace(d.cfg.Settings.Queue.Listen)
 	if address == "" {
 		return nil
 	}
+	resolved, nonLoopback := resolveQueueListenAddress(address)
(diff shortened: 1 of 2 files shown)


orlandoburli:
## Council Verdict: PASS (5/5 APPROVE)

Decision rule satisfied: no Correctness or Security veto; 5/5 lenses approve (threshold >=3).

`resolveQueueListenAddress` handles the cases correctly. `net.SplitHostPort(":8080")` yields host="", port="8080" → `JoinHostPort("127.0.0.1","8080")`. Explicit loopback (`127.0.0.1`, `[::1]`, `localhost`) passes through; non-loopback (`0.0.0.0`, LAN IP) passes through with `nonLoopback=true`. IPv6 unspecified `[::]` correctly warns (IsLoopback false). Error path passes the raw address through unchanged, preserving prior behavior for genuinely malformed input. Table test covers the meaningful branches.

Directly addresses SEC-03: the insecure default (bind-all when only a port is given) now defaults to loopback, shrinking the attack surface of the unauthenticated-by-default worker protocol port. Non-loopback binds remain possible but are now loud (Warn), which is the right posture for an opt-in exposure. TLS is correctly scoped out to SEC-13. No new exposure introduced.

Uses stdlib `net` helpers idiomatically, named returns documented, doc comment present, warning message is actionable and points the operator at the exact fix. Matches surrounding `aplog`/error-wrapping style.

Pure helper, no state, easily unit-tested. Minimal footprint (+56/-2). The error-branch pass-through is the simplest safe fallback.

### Pragmatist — APPROVE
Backward-compatible: existing explicit configs are untouched; only the ambiguous `:port` form changes, toward the safer default. Ships value now, defers TLS cleanly.

### Minor notes (non-blocking)
- The `"::1]:8080"` test case is labeled 'malformed — pass through'; it exercises the SplitHostPort error branch, which is fine, but a comment clarifying that `[::1]:8080` (bracketed) is the well-formed loopback form already covered separately would aid future readers.
- Consider whether `localhost` resolving to a non-loopback entry in an unusual `/etc/hosts` is in-scope; treating it as safe-by-name is a reasonable pragmatic choice.

_Advisory review only — not merging or approving._

PR #258, opened 2026-07-24
feat(queue): native TLS for the worker protocol listener (SEC-13)

orlandoburli:
## Summary

- Adds `tls_cert_file` and `tls_key_file` fields to `settings.queue` so the control-plane listener can terminate TLS natively without a reverse proxy.
- When both fields are set the listener is started with `crypto/tls` (minimum TLS 1.2); omitting both keeps the existing plain-HTTP behaviour.
- `queueControlPlaneURL` now accepts a `tlsEnabled` flag and auto-derives `https://` for workers reading their control-plane URL from config.
- Docs updated to document native TLS (Option A) alongside the existing proxy approach (Option B).

## Changes

| File | Change |
|---|---|
| `src/internal/config/config.go` | `TLSCertFile`, `TLSKeyFile` fields + `TLSEnabled()` on `QueueSettings` |
| `src/internal/daemon/queue_server.go` | Load X.509 key pair; use `tls.Listen` when TLS is configured |
| `src/internal/cli/worker.go` | `queueControlPlaneURL(listen, tlsEnabled)` — emit `https://` when TLS on |
| `src/internal/cli/worker_test.go` | Updated `TestQueueControlPlaneURL` to cover TLS case |
| `src/internal/queuehttp/protocol_test.go` | New `TestWorkerLifecycleOverTLS` using `httptest.NewTLSServer` |
| `docs/distributed-workers.md` | Document Option A (native TLS) and Option B (reverse proxy) |

## Test plan

- [ ] `go test ./internal/cli/... ./internal/queuehttp/...` — all pass including new `TestWorkerLifecycleOverTLS` and updated `TestQueueControlPlaneURL`
- [ ] `go build ./...` — clean build
- [ ] Plain-HTTP path: omitting `tls_cert_file`/`tls_key_file` keeps existing behaviour
- [ ] `--control-plane https://...` continues to work (explicit scheme wins over auto-derive)

Closes #236

PR #276, opened 2026-07-25
feat(queuehttp): add optional TLS for the worker queue transport (SEC-13)

orlandoburli:
## Summary

- Adds `tls_cert_file`, `tls_key_file`, and `tls_ca_file` fields to `QueueSettings` so operators can opt in to TLS without breaking existing plain-HTTP deployments.
- The queue protocol server switches from `http.Server.Serve` to `ServeTLS` when cert+key paths are configured; the log line distinguishes the two modes.
- The worker CLI auto-derives `https://` control-plane URLs when TLS is enabled on the same node, and builds a custom-CA `http.Client` when `tls_ca_file` is set (self-signed cert support).
- Config validation rejects mismatched cert/key pairs and an orphaned `tls_ca_file` without cert+key.
- New `TestTLSWorkerLifecycle` test exercises a full register→claim→finish round-trip through `httptest.NewTLSServer`; existing `TestQueueControlPlaneURL` table is extended to cover the TLS/non-TLS scheme branches.

## Test plan

- [x] `go test -count=1 ./internal/queuehttp/... ./internal/config/... ./internal/daemon/... ./internal/cli/...` — all green
- [x] `TestTLSWorkerLifecycle` — new test verifies end-to-end TLS worker flow
- [x] `TestQueueControlPlaneURL` — extended to assert `https://` scheme when TLS enabled
- [ ] Smoke: run `apiary worker --control-plane https://<host>:PORT` against a daemon with `tls_cert_file`/`tls_key_file` set

Closes #236

### DIFF
--- a/src/internal/cli/worker.go
+++ b/src/internal/cli/worker.go
@@ -2,7 +2,10 @@ package cli
 
 import (
 	"context"
+	"crypto/tls"
+	"crypto/x509"
 	"fmt"
+	"net/http"
 	"os"
 	"os/signal"
 	"path/filepath"
@@ -38,7 +41,8 @@ func newWorkerCmd() *cobra.Command {
 				return fmt.Errorf("config validation failed: %v", errs)
 			}
 			if controlPlane == "" {
-				controlPlane = queueControlPlaneURL(cfg.Settings.Queue.Listen)
+				tlsEnabled := strings.TrimSpace(cfg.Settings.Queue.TLSCertFile) != ""
+				controlPlane = queueControlPlaneURL(cfg.Settings.Queue.Listen, tlsEnabled)
 			}
 			if controlPlane == "" {
 				return fmt.Errorf("--control-plane is required (for example http://apiary-control:8080)")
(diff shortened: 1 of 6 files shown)


orlandoburli (commented):
## Review

PR #276 is the more complete implementation and should be the one merged. PR #258 is a strict subset of this work (same first two commits, no CA-file support) and should be closed as superseded.

### Correctness bug — `tls_ca_file` validation is too restrictive

**File:** `src/internal/config/validate.go`

```go
if strings.TrimSpace(settings.TLSCAFile) != "" && !hasCert {
    errs = append(errs, fmt.Errorf("settings.queue.tls_ca_file requires tls_cert_file and tls_key_file to be set"))
}
```

`tls_ca_file` is a **client-side** trust setting (used by `workerHTTPClient` on the worker node to trust the control-plane's CA). `tls_cert_file`/`tls_key_file` are **server-side** settings (used by the daemon to start the TLS listener). A remote worker running on a separate machine with its own `apiary.yaml` legitimately sets only `tls_ca_file` — it has no cert to serve and no reason to configure `tls_cert_file`/`tls_key_file`. This validation would reject that configuration.

Fix: remove the `!hasCert` guard entirely, or validate `tls_ca_file` independently of cert+key. The cert+key mismatch check (one without the other) is correct and should stay.

### Missing explicit TLS 1.2 minimum

PR #258 set `MinVersion: tls.VersionTLS12` explicitly in `tls.Config`. PR #276 uses `server.ServeTLS` which relies on Go's default (TLS 1.2 since Go 1.18). Consider pinning it explicitly:

```go
if tlsEnabled {
    server.TLSConfig = &tls.Config{MinVersion: tls.VersionTLS12}
    serveErr = server.ServeTLS(listener, certFile, keyFile)
}
```

This makes the security floor visible in the code and is robust against anyone compiling with a very old toolchain.

### `workerHTTPClient` returns the global `http.DefaultClient`

```go
if caFile == "" {
    return http.DefaultClient, nil
}
```

`http.DefaultClient` is a global. Any mutation to it elsewhere (headers, transport, timeout) would silently affect queue worker connections. Prefer returning `&http.Client{}` (which uses `http.DefaultTransport` via the zero value, giving identical behaviour but no aliasing risk).

### No test for the CA-file code path

`TestTLSWorkerLifecycle` uses `httptest.NewTLSServer.Client()` directly, which bypasses `workerHTTPClient`. The `pool.AppendCertsFromPEM` call and the `http.Transport{TLSClientConfig: &tls.Config{RootCAs: pool}}` construction are untested. A unit test for `workerHTTPClient` with a temp PEM file would give coverage with low effort.

---

**Overall:** the core TLS listener and `ServeTLS` integration are correct. The CA-file validation bug is the only change required before merge; the rest are hardening suggestions.

orlandoburli (commented):
review posted

PR #279, opened 2026-07-25
security: check per-plugin subdirectory permissions in Discover (SEC-04 follow-up)

orlandoburli:
## Summary

- Extends the plugin directory permission check from #241 to cover the standard multi-plugin layout (`plugins/<a>/`, `plugins/<b>/`).
- Before this fix, `checkDirOwnerOnly` was called only on the configured top-level path; each per-plugin subdirectory entered the load loop unchecked. A correctly-locked parent containing a group/world-writable child still served any executable planted there.
- Fix: inside the roots loop, call `checkDirOwnerOnly(root)` for every root that is a subdirectory (`root != path`); skip and warn on failure, mirroring the existing top-level guard.
- Test: splits `TestDiscoverSkipsGroupAndWorldWritableDirs` into two named sub-tests and adds a new one — owner-only parent with one writable child and one safe sibling — verifying the writable child is skipped and the safe sibling loads.

## Test plan

- [x] `go test ./internal/plugin/...` passes (both sub-tests green)
- [x] New sub-test `writable child plugin dir is skipped, sibling is loaded` fails before the fix and passes after
- [x] Existing `writable parent dir is refused entirely` behaviour unchanged

Closes #245

### DIFF
--- a/src/internal/plugin/discovery.go
+++ b/src/internal/plugin/discovery.go
@@ -53,6 +53,15 @@ func Discover(paths []string, baseDir, apiaryVersion string) (*Registry, []error
 		}
 		sort.Strings(roots)
 		for _, root := range roots {
+			if root != path {
+				// Per-plugin subdirectory: enforce owner-only permissions just like
+				// the parent. A writable child dir is the common attack surface —
+				// it bypasses the top-level check when the parent is correctly locked.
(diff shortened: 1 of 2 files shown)


orlandoburli:
Closing as duplicate of #266, which has the same fix with a cleaner unconditional check. PR #266 has the full test coverage for the writable-child case and is already open for review.

PR #252, opened 2026-07-24
security: wrap untrusted issue content in delimited prompt block (SEC-06)

orlandoburli:
## Summary

- All ticket-derived fields (title, type, priority, labels, URL, description) inside `buildPrompt` (`runner/execution/cli.go`) are now enclosed in `<untrusted-content>…</untrusted-content>` tags.
- A preamble line inside the block explicitly instructs the LLM not to treat the enclosed text as instructions.
- System-controlled fields (`SystemPrepend`, `SystemAppend`, `OutputInstruction`, `SummaryPrompt`) remain outside the delimiter and are unaffected.

## Test plan

- [ ] `go build ./...` passes (verified locally).
- [ ] Run an agent task and confirm the prompt sent to the LLM contains the `<untrusted-content>` wrapper around the issue fields.
- [ ] Verify system prepend/append text still appears outside the delimiter.

Closes #229

PR #288, opened 2026-07-25
security: wrap untrusted ticket fields in delimited prompt block (SEC-06)

### LINKED ISSUE #291: [security] Rework SEC-06: strip closing delimiter from ticket fields (PR #252 rejected)
Council **rejected** PR #252 (Security veto). The delimiter-wrapping is trivially bypassable.

**Defect:** the closing `</untrusted-content>` marker is never stripped/escaped from attacker-controlled fields (esp. `Description`), so a payload can emit the closing tag and break out of the untrusted block ahead of the trusted system/output instructions.

**Fix direction:** strip/escape the delimiter tokens from all ticket-derived fields before wrapping (or use an unguessable per-run nonce delimiter). Add a test with a payload containing the closing tag.
Ref PR #252 (SEC-06).

orlandoburli:
## Summary

- Ticket title, type, priority, labels, URL, and description were concatenated directly into the agent prompt with no boundary markers, allowing a malicious ticket to inject instructions that override the agent's system prompt (original SEC-06 wrapping, PR #252 context)
- Wraps all externally-sourced ticket fields inside an explicit \`=== BEGIN UNTRUSTED TICKET CONTENT ... === END UNTRUSTED TICKET CONTENT ===\` delimiter pair in \`buildPrompt\`
- **Rework (PR #252 rejected — SEC-06):** adds \`sanitizeTicketField\` which strips both the opening and closing delimiter tokens from all ticket-derived fields before wrapping, preventing a crafted payload from closing the untrusted block early and injecting trusted instructions
- Both \`CliRunner\` and \`ApiRunner\` share the same \`buildPrompt\` helper (\`src/internal/runner/execution/cli.go\`), so a single change covers all execution paths
- Adds \`TestBuildPrompt_DelimiterStrippedFromFields\` with a bypass payload containing \`=== END UNTRUSTED TICKET CONTENT ===\` and asserts the closing delimiter appears exactly once (the one we write ourselves)

## Test plan

- [x] \`go test ./internal/runner/execution/...\` passes locally — including \`TestBuildPrompt_DelimiterStrippedFromFields\`
- [x] GitNexus detect_changes confirms only \`buildPrompt\`, \`sanitizeTicketField\`, and their tests are touched
- [ ] CI gate passes

Closes #229
Closes #291

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,25 +1,24 @@
 <!-- gitnexus:start -->
 # GitNexus — Code Intelligence
 
-This project is indexed by GitNexus as **apiary** (6306 symbols, 22913 relationships, 300 execution flows). Use the GitNexus MCP tools to understand code, assess impact, and navigate safely.
+This project is indexed by GitNexus as **apiary** (10270 symbols, 23355 relationships, 300 execution flows). Use the GitNexus MCP tools to understand code, assess impact, and navigate safely.
 
-> Index stale? Run `node .gitnexus/run.cjs analyze` from the project root — it auto-selects an available runner. No `.gitnexus/run.cjs` yet? `npx gitnexus analyze` (npm 11 crash → `npm i -g gitnexus`; #1939).
+> If any GitNexus tool warns the index is stale, run `npx gitnexus analyze` in terminal first.
 
 ## Always Do
 
-- **MUST run impact analysis before editing any symbol.** Before modifying a function, class, or method, run `impact({target: "symbolName", direction: "upstream"})` and report the blast radius (direct callers, affected processes, risk level) to the user.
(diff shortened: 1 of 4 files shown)


orlandoburli:
Rework addressing the council's security veto on PR #252.

**Root cause fixed:** `sanitizeTicketField` now strips both `untrustedOpen` and `untrustedClose` from every ticket-derived field (Title, Type, Priority, Labels, URL, Description) before they are embedded in the prompt. Without this, a description containing `=== END UNTRUSTED TICKET CONTENT ===` would close the sandboxed block early and allow injected text to reach the trusted instruction zone.

**New regression test:** `TestBuildPrompt_DelimiterStrippedFromFields` embeds the closing delimiter in every ticket field and asserts the delimiter appears exactly once in the final prompt (the one the system writes itself).

Closes #291

PR #295, opened 2026-07-25
security: sandbox agent execution — filesystem/process containment (SEC-09) [#289]

orlandoburli:
## Summary
Runs each agent subprocess inside a Docker container, isolated from the host filesystem. Reworked after council review of #254 and of this PR.

## What this actually contains — and what it does not
**Contains:** host filesystem access (only the task working directory is mounted), process/capability escalation (`--read-only`, `--cap-drop all`, `--security-opt no-new-privileges`, tmpfs `/tmp`), and unrelated host secrets (env is allow-listed).

**Does NOT contain: exfiltration.** `network` defaults to `bridge` because coding agents must reach their LLM API and git remotes, and the agent legitimately holds provider keys and a `GITHUB_TOKEN`. An injected agent can still send data out. Egress policy is the missing piece and is tracked as follow-up on #289.

That is why this is `Refs #289`, **not `Closes`** — it narrows the blast radius of a prompt injection but does not shut the rollout gate on its own.

## Fixes from review of #254
- **Provider keys retained.** `scopedEnv` allow-lists system vars plus LLM/agent provider families (`ANTHROPIC_`, `OPENAI_`, `OPENCODE_`, `CURSOR_`, …) by prefix, plus operator-configurable `env_passthrough`. Unrelated host secrets (`AWS_*` etc.) are dropped. #254 stripped provider keys and would have broken every run.
- **No secret values in argv.** Credentials cross with the name-only form `docker run --env NAME`; values live only in the docker process's own env, never the host process table.

## Fixes from review of this PR
- **`--user` defaults to the daemon's uid:gid, not `nobody`.** The workspace is bind-mounted and owned by the daemon user, so `nobody` made every write fail with EACCES — the sandbox was unusable for a coding agent.
- **`DOCKER_HOST` is no longer forwarded into the container** — forwarding the host Docker socket pointer from an isolation feature is a trivial host escape, a hole this PR itself introduced. `SSH_AUTH_SOCK` and host paths/identity (`HOME`/`PATH`/`TMPDIR`/`XDG_*`) are filtered too; `HOME` under a read-only rootfs was also a functional break.
- **Fails closed on MCP + sandbox.** `setupMCP` writes provider config to host paths (`/tmp`, `~/.cursor`, `~/.config`) that the container cannot see (and `/tmp` is masked by tmpfs), so the agent would have started with MCP servers silently missing. Configure now errors instead.
- **`extra_args` is validated** — `--privileged`, `--cap-add`, `--user`, `--network`, extra mounts and friends are refused rather than silently undoing the sandbox.

## Known gaps (deliberate, not oversights)
- Egress/exfiltration, as above.
- MCP is unsupported under a sandbox (fails closed rather than degrading silently).
- Opt-in per runner via `runners[].sandbox`; nothing changes for runners without it.

## Tests
Provider keys retained / unrelated secrets dropped / overlay precedence; no `NAME=VALUE` in argv; host wiring not forwarded while credentials are; writable-workspace uid; `extra_args` rejection. `go build`, `go vet`, and the `execution` + `config` + `daemon` package tests pass.

Refs #289

### DIFF
--- a/docs/runners.md
+++ b/docs/runners.md
@@ -185,6 +185,56 @@ pre-authenticated CLI tool exists on the host.
     capable for implementation work — prefer `cli` runners for coding agents
     and API runners for lightweight classification or as fallback capacity.
 
+## Sandboxing agent execution
+
+A CLI runner can run every agent subprocess inside a Docker container, isolating
+it from the host filesystem. Use it for runners that process issues or comments
+written by people you do not fully trust.
+
+```yaml
+runners:
+  - id: sandboxed-claude
+    type: cli
+    provider: claude
+    config:
+      command: claude
+    sandbox:
+      image: my-org/apiary-agent:latest   # must contain the agent binary
+      # user:    "1000:1000"              # default: the daemon's uid:gid
+      # network: none                     # default: bridge (agents need egress)
+      extra_args: ["--memory", "4g", "--pids-limit", "512"]   # resource limits only
+    env_passthrough: ["MYCORP_*"]         # beyond system + provider credentials
+```
+
+**What it contains:** host filesystem access (only the task working directory is
+mounted), process and capability escalation (`--read-only` rootfs, `--cap-drop
+all`, `--security-opt no-new-privileges`), and unrelated host secrets — the agent
+environment is allow-listed to system variables plus LLM provider credentials,
+with per-task credentials overlaid. Credentials are passed to the container by
+name, so their values never appear in the host process table.
+
+**What it does NOT contain: network exfiltration.** `network` defaults to
+`bridge` because coding agents must reach their LLM API and git remotes, and the
+agent legitimately holds provider keys and a source token. A prompt-injected
+agent can still send data out. Add egress controls at the network layer if your
+trust level requires it.
+
+Notes:
+
+- `extra_args` accepts only resource-limit and labelling flags (`--memory`,
+  `--cpus`, `--pids-limit`, `--ulimit`, `--label`, …). Anything that could weaken
+  the sandbox — `--privileged`, `--cap-add`, extra mounts, `--network`,
+  `--user`, `--read-only=false`, `--entrypoint` — is rejected at config load.
+- `HOME` inside the container points at a writable tmpfs (`mode=1777`, `exec`),
+  since the rootfs is read-only and a numeric `--user` has no passwd entry. Both
+  options are set explicitly because docker *merges* its tmpfs defaults
+  (`nodev,noexec,relatime`) rather than replacing them, and a tmpfs otherwise
+  inherits the mountpoint's mode from the image.
+- **MCP servers are not supported with a sandbox yet.** Their config is written to
+  host paths the container cannot see, so combining them is rejected at config
+  load rather than silently starting an agent with its MCP servers missing.
+
+
 ## MCP servers
 
 Runners (and individual agents) can expose
--- a/schema/apiary.json
+++ b/schema/apiary.json
@@ -20,6 +20,27 @@
         "additionalProperties": false,
         "required": ["id", "type"],
         "properties": {
+          "sandbox": {
+            "type": "object",
+            "description": "Run this runner's agent subprocesses inside a Docker container isolated from the host filesystem. Contains filesystem/process access; does NOT contain network egress.",
+            "additionalProperties": false,
+            "required": ["image"],
+            "properties": {
+              "image": {"type": "string", "description": "Docker image providing the agent binary"},
+              "user": {"type": "string", "description": "--user value; defaults to the daemon's uid:gid so the bind-mounted workspace stays writable"},
+              "network": {"type": "string", "description": "--network value; defaults to bridge because agents need LLM/git egress. Use \"none\" for agents needing no network"},
+              "extra_args": {
+                "type": "array",
+                "items": {"type": "string"},
+                "description": "Additional docker flags. Restricted to resource-limit and labelling flags (e.g. --memory, --cpus, --pids-limit, --ulimit, --label); anything that could weaken the sandbox is rejected"
+              }
+            }
+          },
+          "env_passthrough": {
+            "type": "array",
+            "items": {"type": "string"},
+            "description": "Extra host environment variable names to forward to agents beyond the built-in allow-list (system vars + LLM provider credentials). Exact names or a trailing-* prefix, e.g. MYCORP_*"
+          },
           "id": {
             "type": "string",
             "description": "Unique runner identifier, referenced by agents[].runner and agents[].fallbacks"
--- a/src/internal/config/config.go
+++ b/src/internal/config/config.go
@@ -73,6 +73,39 @@ type RunnerConfig struct {
 	// format/location (see runner/execution). Agent-scope MCPs (AgentConfig.MCPs)
 	// are layered on top of these, overriding by name.
 	MCPs []model.MCPServer `yaml:"mcps,omitempty"`
+	// Sandbox, when set, wraps every agent subprocess for this runner in a Docker
+	// container that isolates it from the host filesystem (only the task working
+	// directory is mounted) and runs it as an unprivileged user. Use it for
+	// runners that process issues/comments from external or untrusted authors, to
+	// contain a successful prompt injection.
+	Sandbox *SandboxConfig `yaml:"sandbox,omitempty"`
+	// EnvPassthrough lists additional host environment variable names to forward to
+	// the agent subprocess beyond the built-in allowlist (system vars + LLM/agent
+	// provider credentials). Entries are exact names or a trailing-"*" prefix
+	// (e.g. "MYCORP_*"). Host variables not covered by the allowlist or this list
+	// are never inherited, so unrelated daemon secrets cannot leak into an agent.
+	EnvPassthrough []string `yaml:"env_passthrough,omitempty"`
+}
+
+// SandboxConfig describes Docker container isolation for a CLI runner's agent
+// subprocesses. Network is left enabled by default because coding agents must
+// reach their LLM API and git remotes; set Network to "none" only for agents
+// that need no network.
+type SandboxConfig struct {
+	// Image is the Docker image providing the agent binary (required).
+	Image string `yaml:"image"`
+	// User is passed as --user (e.g. "1000:1000"). Defaults to the daemon's own
+	// uid:gid so the bind-mounted workspace stays writable — "nobody" would make
+	// every workspace write fail with EACCES.
+	User string `yaml:"user,omitempty"`
+	// Network is the --network value (e.g. "bridge", "none"). Default "bridge".
+	Network string `yaml:"network,omitempty"`
+	// ExtraArgs are appended after the docker flags and before the image. Only
+	// resource-limit and labelling flags are permitted (e.g.
+	// ["--memory", "4g", "--pids-limit", "512"]); anything that could weaken the
+	// sandbox — extra mounts, --privileged, --cap-add, --user, --network,
+	// --read-only=false, --entrypoint — is rejected at config load.
+	ExtraArgs []string `yaml:"extra_args,omitempty"`
 }
 
 // AdapterName returns the runner adapter name as "{provider}-{type}" when both
--- a/src/internal/daemon/dispatcher.go
+++ b/src/internal/daemon/dispatcher.go
@@ -357,7 +357,7 @@ func New(ctx context.Context, cfg *config.Config, configFile string, dbClient *d
 			return nil, fmt.Errorf("agent %q: runner type %q not found", ac.ID, rc.Type)
 		}
 
-		if err := ra.Configure(runnerConfigWithMCPs(rc.Config, rc.MCPs, ac.MCPs)); err != nil {
+		if err := ra.Configure(injectRunnerSecurity(runnerConfigWithMCPs(rc.Config, rc.MCPs, ac.MCPs), rc.Sandbox, rc.EnvPassthrough)); err != nil {
 			return nil, fmt.Errorf("agent %q: configure runner: %w", ac.ID, err)
 		}
 
@@ -389,7 +389,7 @@ func New(ctx context.Context, cfg *config.Config, configFile string, dbClient *d
 			if !ok {
 				return nil, fmt.Errorf("agent %q: fallback runner type %q not found", ac.ID, fAdapterName)
 			}
-			if err := fra.Configure(runnerConfigWithMCPs(frc.Config, frc.MCPs, ac.MCPs)); err != nil {
+			if err := fra.Configure(injectRunnerSecurity(runnerConfigWithMCPs(frc.Config, frc.MCPs, ac.MCPs), frc.Sandbox, frc.EnvPassthrough)); err != nil {
 				return nil, fmt.Errorf("agent %q: configure fallback runner %q: %w", ac.ID, fb.Runner, err)
 			}
 			if fAdapterName == "opencode" {
@@ -1845,6 +1845,39 @@ func runnerConfigWithMCPs(base map[string]any, runnerMCPs, agentMCPs []model.MCP
 	return out
 }
 
+// injectRunnerSecurity folds a runner's sandbox and env-passthrough settings into
+// the config map handed to the runner's Configure. The base map is never mutated;
+// a copy is returned only when there is something to add.
+func injectRunnerSecurity(base map[string]any, sandbox *config.SandboxConfig, envPassthrough []string) map[string]any {
+	if sandbox == nil && len(envPassthrough) == 0 {
+		return base
+	}
+	out := make(map[string]any, len(base)+2)
+	for k, v := range base {
+		out[k] = v
+	}
+	if sandbox != nil {
+		extra := make([]any, len(sandbox.ExtraArgs))
+		for i, v := range sandbox.ExtraArgs {
+			extra[i] = v
+		}
+		out["sandbox"] = map[string]any{
+			"image":      sandbox.Image,
+			"user":       sandbox.User,
+			"network":    sandbox.Network,
+			"extra_args": extra,
+		}
+	}
+	if len(envPassthrough) > 0 {
+		p := make([]any, len(envPassthrough))
+		for i, v := range envPassthrough {
+			p[i] = v
+		}
+		out["env_passthrough"] = p
+	}
+	return out
+}
+
 func workerRunConfig(wc config.WorkerConfig) map[string]any {
 	m := map[string]any{}
 	if wc.RunnerConfig != nil {
@@ -1891,7 +1924,7 @@ func (d *Dispatcher) UpdateAgentConfig(ctx context.Context, agentID, newModel, n
 		if !ok {
 			return fmt.Errorf("runner type %q not registered", rc.Type)
 		}
-		if err := ra.Configure(rc.Config); err != nil {
+		if err := ra.Configure(injectRunnerSecurity(rc.Config, rc.Sandbox, rc.EnvPassthrough)); err != nil {
 			return fmt.Errorf("configure runner %q: %w", newRunner, err)
 		}
 
--- a/src/internal/dashboard/render_test.go
+++ b/src/internal/dashboard/render_test.go
@@ -541,11 +541,11 @@ func TestTaskDetailShowsUsageWhenZero(t *testing.T) {
 	a := newTestApp(80, 24)
 	a.model.tasksTab.View = TaskViewDetail
 	a.model.tasksTab.Detail = &TaskItem{
-		TaskID:  "T-2",
-		Number:  "PRJ-6",
-		Title:   "old task",
-		Agent:   "engineer",
-		Status:  "success",
+		TaskID: "T-2",
+		Number: "PRJ-6",
+		Title:  "old task",
+		Agent:  "engineer",
+		Status: "success",
 	}
 	out := stripANSI(a.renderTaskDetail(a.model.tasksTab, 20))
 	if !strings.Contains(out, "0 in / 0 out / 0 total") {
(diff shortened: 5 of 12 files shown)


orlandoburli:
## 🐝 Apiary Review Council — PR #295 (SEC-09 rework of #254)

Verified both original vetoes, then hunted for new defects introduced by the rework.

**Veto (a) — provider keys:** genuinely addressed. `providerEnvPrefixes` covers `ANTHROPIC_`, `CLAUDE_`, `OPENAI_`, `OPENCODE_`, `CURSOR_`, `GEMINI_`, etc. `scopedEnv` retains them and the overlay wins on collision. ✅
**Veto (b) — secrets in argv:** genuinely fixed. `wrapCommand` emits name-only `--env NAME`; values live only in `cmd.Env` of the docker process. ✅

But the rework introduces a new instance of the *same class* of defect that sank #254: the sandbox silently strips capabilities the agent needs.

---

### 1. Correctness Hawk (veto power) — **REJECT**

**Blocker A — MCP servers are silently lost inside the sandbox.** `Configure` runs `setupMCP()` (`cli_mcp.go`), which writes provider config to **host** paths and returns host paths in `mcpRunArgs`:
- claude → `os.CreateTemp("", "apiary-mcp-*.json")` → `--mcp-config /tmp/apiary-mcp-XXXX.json`
- cursor → `$HOME/.cursor/mcp.json`; opencode → `$HOME/.config/opencode/opencode.json`

`wrapCommand` mounts **only** `workDir`. Worse, `--tmpfs /tmp` guarantees the claude MCP file is *not* visible — the container gets an empty tmpfs at exactly that path. So every sandboxed run loses all MCP servers, with no error. Same failure shape as #254 stripping provider keys, just one layer over.

**Blocker B — `HOME` is forwarded as a host path that does not exist in the image.** `HOME` is in `systemEnvAllow`, so the container receives e.g. `HOME=/home/apiary` with nothing mounted there and `--read-only` rootfs. Agent CLIs that read/write `$HOME` (`~/.claude/.credentials.json`, `~/.config/opencode`) fail. Note this also means **Claude Code OAuth/subscription auth has no path into the container at all** — the "provider keys retained" fix only covers API-key-in-env auth, not the credentials-file auth many installs use.

**Blocker C — `--user nobody` + `-v workDir:workDir` cannot write.** The task workdir is owned by the daemon uid; uid 65534 gets EACCES on every file write. There is no uid mapping and no ownership note. An agent that cannot write to its own workspace is a no-op run.

None of these are covered by the tests, which only assert argv shape and never exercise a real container.

### 2. Security Skeptic (veto power) — **REJECT**

The two original findings are fixed, credit where due. But:

**S1 — `DOCKER_HOST` is in `systemEnvAllow`.** An isolation feature forwards, by default, the variable that points at the host Docker daemon. If the operator ever adds a socket mount via `extra_args` (or uses a TCP `DOCKER_HOST`), a prompt-injected agent gets host-root-equivalent escape straight out of its own sandbox. `SSH_AUTH_SOCK` is likewise allowlisted (dead path today, live the moment someone mounts it). Both should be dropped from the default allowlist and opted in via `env_passthrough`.

**S2 — containment goal is not met for the stated threat.** `network` defaults to `bridge` and the container holds `ANTHROPIC_API_KEY` + the scoped `GITHUB_TOKEN`. A successful prompt injection — the thing #232/#289 asks us to contain — can still `curl` every credential out. This PR contains *filesystem* reach only. That is a legitimate scope decision, but the PR title and issue closure ("Closes #232 / #289") overstate what shipped; egress policy is the load-bearing half.

**S3 — `extra_args` is spliced unvalidated ahead of the image**, so `["-v","/:/host"]`, `--privileged`, or `--cap-add` silently undo every hardening flag with no lint or warning. Operator-controlled, but an isolation primitive should refuse to disarm itself quietly.

### 3. Conventions Stickler — **REJECT**

`SandboxConfig` / `EnvPassthrough` are added to `RunnerConfig` with **no `schema/apiary.json` mirror, no `apiary validate` lint, no docs page, and no `.apiary/example-*.yaml` entry** — all four are established requirements in this repo for a new config surface. `sandbox.image` is validated only deep inside `CliRunner.Configure` at daemon-construction time rather than in config validation. Also, `injectRunnerSecurity` reconstructs the typed struct into a `map[string]any` only to re-parse it with untyped assertions in `Configure` — a typo in either half fails silently rather than at compile time.

### 4. Simplicity Advocate — **APPROVE**

`scopedEnv` is clean and correct: `strings.Cut` guards a missing `=`, overlay applied after host so it wins, names sorted and derived from the same map so `--env NAME` can never reference an unset var. No nil/empty-slice or ordering bug found — `wrapCommand` handles `workDir == ""` and `envNames == nil` fine. Sandbox logic lives in its own file, and the non-sandbox branch is byte-for-byte the old behavior (`cmd.Dir`, `os.Environ()`, overlay append) — no silent change for existing users. Nit: the sandbox branch never sets `cmd.Dir`, harmless only because `-w` is absolute.

### 5. Pragmatist — **APPROVE**

Opt-in, defaults off, existing deployments untouched; `go build ./...` and the `execution` tests pass locally. The two review-blocking bugs from #254 are really gone. Shape of the solution is right.

---

## Final decision: ❌ **REJECTED** (Correctness Hawk veto + Security Skeptic veto; 2/5 approve)

**Required before re-review:**
1. Mount the MCP config into the container (or write it under `workDir`) so sandboxed runs keep their MCP servers — with a test that fails if `mcpRunArgs` references a path outside the mount set.
2. Provide a container-local `HOME` (and a credentials path) instead of leaking the host `HOME` value; state explicitly whether OAuth-file auth is supported.
3. Resolve the `--user nobody` vs. workdir-ownership conflict (uid mapping / documented `user: "$(id -u):$(id -g)"` default).
4. Drop `DOCKER_HOST` and `SSH_AUTH_SOCK` from `systemEnvAllow`.
5. Add schema mirror + `apiary validate` lint (including a warning when `extra_args` re-grants dropped privileges) + docs.
6. Do not close #232/#289 on this PR — egress containment is still outstanding.

orlandoburli:
## Reviewer council — re-review of #295 @ f02ff12

**Verdict: REJECTED** (Correctness Hawk: REJECT, Security Skeptic: REJECT, Conventions: REJECT, Simplicity: APPROVE, Pragmatist: APPROVE)

Real progress since the last pass — four of the seven findings are genuinely closed. Two are not, and one of them is the same functional break wearing a different hat.

### Verified fixed
- **(4) DOCKER_HOST / SSH_AUTH_SOCK.** Traced and correct. `systemEnvAllow` still admits them into `cmd.Env`, which is what the **docker client process** needs to reach the daemon; `containerEnvNames` strips them from the `--env NAME` list so they never cross into the container. The split is the right one and `TestWrapCommand_DoesNotForwardHostWiringIntoContainer` pins it.
- **(3) `--user`.** No longer `nobody`; `os.Getuid():os.Getgid()` matches the bind-mounted workspace owner.
- **(1) MCP fail-closed.** Placement is correct and I checked the ordering concern specifically: `Configure` reads keys in *code* order, not map order — `mcps` at L74 precedes the `sandbox` block at L92, and `r.mcps` is a struct field that survives the two-phase configure (`providers.go` factory call, then `dispatcher.go:360/392` with the merged runner+agent MCPs). Agent-level `mcps` added on top of a sandboxed runner still hits the error. Good.
- **(7) Scope honesty.** The body is now accurate: "Does NOT contain: exfiltration" is stated plainly up top, `Refs #289` not `Closes`. This is the standard we asked for.

### Blocking — Correctness Hawk (veto)

**There is still no writable `HOME` inside the container, and the primary supported runner cannot start without one.**

Finding (2) was "HOME points at a nonexistent host path under `--read-only`". The fix removes `HOME` from what crosses the boundary. That resolves the *wrong path* half and leaves the *read-only* half exactly where it was:

- `--user <uid>:<gid>` with a uid that has no `/etc/passwd` entry in the image, and no `HOME` in the env, means docker gives the process `HOME=/`.
- `/` is `--read-only`. The only writable paths in the container are the 64 MB `/tmp` tmpfs and the workspace bind mount.
- `claude` writes `~/.claude.json` and `~/.claude/` on startup. `opencode` and `cursor` likewise want a config/state dir. All of those land on the read-only rootfs and fail.

The code comment concedes the mechanism — *"HOME under a read-only rootfs is the classic failure"* — and then ships without a writable home. There is no `--tmpfs` for a home dir, no volume, no `HOME=` override, no `--env HOME`.

Compounding it: **nothing in this PR has ever executed `docker`.** All eight tests are pure argv-shape assertions on `wrapCommand`. For a containment feature whose entire value is runtime behaviour, "`go vet` and the unit tests pass" is not evidence that a single agent run succeeds. Please run one real sandboxed `claude-cli` task end to end and paste the result; if it works, my objection is wrong and I will withdraw it, but the structural argument above says it will not.

### Blocking — Security Skeptic (veto)

**`validateExtraArgs` is a name-denylist over a flag surface docker parses far more liberally, and it omits the flag that most directly undoes this PR.**

- `--read-only=false` is **not** in `dangerousDockerArgs`. `extraArgs` are appended after the hardening flags, and docker is last-wins, so a single benign-looking entry silently disables the headline containment. Same for `--entrypoint`, `--cgroupns=host`, `--group-add`, and `--tmpfs /etc` (writable overlay on a rootfs path).
- Attached-shorthand bypasses the mount and user checks entirely. Docker uses spf13/pflag, which accepts `-v/:/hostfs` and `-u0:0` as single tokens — no `=`, no separate arg, so `flag` never equals `-v` or `-u` and both sail through. The test only covers the separated `{"-v", "/:/host"}` form.

I accept the PR's framing that `extra_args` is operator-owned rather than attacker-controlled — that is why this is a "fix it", not "the sky is falling". But a guard that *looks* exhaustive and misses `--read-only=false` is worse than no guard, because it invites operators to trust it. Either flip to an allowlist of known-benign flags (`--memory`, `--cpus`, `--pids-limit`, `--dns`, …) or normalise the token before matching and document the denylist as best-effort.

### Blocking — Conventions Stickler

Finding (6) is unfixed, and it is worse than "just missing docs" here. `schema/apiary.json` has `runners[].items.additionalProperties: false`, and neither `sandbox` nor `env_passthrough` was added. Any operator who writes the config this PR introduces gets a **schema error in their editor** for a feature that does not exist in `schema/apiary.json`, `docs/runners.md`, `docs/configuration.md`, or `.apiary/example-apiary-full.yaml`. House rule is to re-mirror the schema and examples from the Go structs on every struct change. On its own this would not veto an opt-in feature; combined with the MCP incompatibility — which an operator can only discover by restarting the daemon and reading a Go error string — the feature is effectively undiscoverable and undocumented at the exact points where it fails closed.

### Non-blocking, but fix while you are in here
- `/tmp` is mounted `noexec`. npm/npx, node native-module extraction, and git hooks commonly exec out of `/tmp`; expect breakage even once HOME is solved.
- `emit("debug", "$ %s %s", r.command, argv)` logs the **unwrapped** command. In a security feature, the logs should show what actually ran — right now you cannot tell from a transcript whether the sandbox applied.
- `cliSandbox.user`'s doc comment still says `// --user; default "nobody"`. Stale as of this commit.

### To unblock
1. Give the container a writable home (`--tmpfs /home/agent:rw,size=…` plus `--env HOME=/home/agent`, or a dedicated state volume) **and** post evidence of one real sandboxed run.
2. Close the `extra_args` bypasses (`--read-only=false`, attached shorthand) or switch to an allowlist.
3. Mirror `sandbox` / `env_passthrough` into `schema/apiary.json`, `docs/runners.md`, and an example yaml, documenting the MCP incompatibility where an operator will read it before hitting it.

`gofmt -l` clean on touched files, `go vet` clean, `execution` + `config` + `daemon` tests pass.

orlandoburli:
## Council re-review of 0c4037d — REJECTED (Correctness veto)

Real progress: the tests now genuinely run containers (all five pass here against Docker 29.6.2), `extra_args` is a sound allow-list, the commit message is honest about exfiltration and correctly says `Refs #289`. But **both of the two claimed runtime fixes are wrong, and I proved it with docker**.

### 1. `noexec` is still set on both tmpfs mounts (Correctness — VETO)
The commit message says: *"`/tmp` is not noexec, because agents legitimately execute from it (npx, node native modules, git hooks)."* That is false as shipped. Docker **merges** its tmpfs defaults (`noexec,nodev`) with the options you supply — it does not replace them. Running the exact flags `wrapCommand` emits:

```
tmpfs on /tmp        type tmpfs (rw,nosuid,nodev,noexec,relatime,size=524288k)
tmpfs on /home/apiary type tmpfs (rw,nosuid,nodev,noexec,relatime,size=524288k)
$ cp /bin/busybox /tmp/bb && /tmp/bb echo   ->  Permission denied
```

So the previously-raised `noexec` defect is not fixed, and now it also applies to `HOME` — where `~/.npm/_npx`, node native `.node` modules, and CLI plugins/hooks live. Fix: add `exec` explicitly. Verified working:
`--tmpfs /tmp:rw,exec,nosuid,size=512m` -> `tmpfs on /tmp type tmpfs (rw,nosuid,nodev,relatime,...)`.

### 2. The writable-`HOME` fix breaks on any image that pre-creates `/home/apiary` (Correctness — VETO)
A tmpfs inherits the **mountpoint directory's** mode. `busybox` has no `/home/apiary`, so docker creates it `1777` and the test passes. A purpose-built agent image (which the docs require — *"must contain the agent binary"*) very plausibly has a home dir there. Then:

```
# image: RUN mkdir -p /home/apiary && chown 1000:1000 /home/apiary   (or 0700 root)
$ docker run --read-only --user 501:20 --tmpfs /home/apiary:rw,nosuid,size=512m --env HOME=/home/apiary IMG \
    sh -c 'ls -ld /home/apiary; touch $HOME/x || echo HOMEWRITE_FAIL'
drwxr-xr-x  2 root root  /home/apiary
touch: /home/apiary/x: Permission denied
HOMEWRITE_FAIL
```

This is last round's veto moved from `/` to `/home/apiary`, not closed. `TestSandbox_HomeIsWritable` cannot catch it because `busybox` is the one image where it happens to work — **the test passes with a broken sandbox**. Fix (verified): add `mode=1777` to the tmpfs options. Please also add a docker test using an image that pre-creates the dir.

### 3. `extra_args` — no bypass found (Security: PASS)
I ran ~40 hostile inputs through `validateExtraArgs`: `--read-only=false`, `--read-only false`, `-v/:/hostfs`, `-u0:0`, `--entrypoint`, `--cgroupns=host`, `--group-add`, `--tmpfs /etc`, `--pid=host`, `--userns=host`, `--cap-add=SYS_ADMIN`, `--security-opt seccomp=unconfined`, `--device`, `--mount`, `--volumes-from`, `--sysctl`, `---privileged`, `--`, `""`, `-`, `--PRIVILEGED`, unicode-hyphen `--‐privileged`, `-m4g`, trailing-flag, and every "value swallows a denied token" form. **All rejected.** I also confirmed against real docker that the swallow cases are not exploitable — `docker run --label --privileged busybox …` runs *unprivileged* with the literal label; `--pull/--platform/--dns/--name` all hard-error on the token. The value/bool split matches docker's actual flag types in both directions. Good work.
Nit: `--dns`, `--dns-search`, `--dns-option` are neither resource limits nor labelling, contrary to the doc/error text — operator-trust, so not blocking.

### 4. Stale/false documentation of shipped behaviour (Conventions — REJECT)
`src/internal/config/config.go` still contradicts the code and the commit message:
- L97 `// User is passed as --user (e.g. "nobody", "1000:1000"). Default "nobody".` — the default is the daemon uid:gid. The "stale nobody comment" fix landed in `sandbox.go` only.
- L102 `// ExtraArgs … e.g. ["-v", "/cache:/cache:ro"] for an additional read-only mount.` — that example is now **rejected at load**.
- `docs/runners.md`: *"`HOME` inside the container points at a writable tmpfs"* — conditionally false per (2).
- No `.apiary/example-*.yaml` entry for `sandbox`/`env_passthrough`.

### Minor
- `TestSandbox_HostWiringNotForwarded` is near-tautological (`containerEnvNames` never hands docker the names, so nothing could leak) and sets an unused `DOCKER_HOST_PROBE`.
- The MCP fail-closed check runs *after* `setupMCP` already wrote host MCP config / merged cursor+opencode global config. Cosmetic (load aborts anyway), but move it above.

### Lenses
- **Correctness Hawk: REJECT** (veto) — both runtime fixes are unverified assumptions about docker tmpfs semantics; both are wrong.
- **Security Skeptic: APPROVE** — allow-list holds under attack, env scoping and name-only credential delivery are sound, scope honesty is exemplary.
- **Conventions Stickler: REJECT** — public config type documents behaviour the code rejects.
- **Simplicity Advocate: APPROVE** — allow-list is far easier to reason about than the deny-list.
- **Pragmatist: REJECT** — the fix is two option strings, but as shipped a realistic agent image fails on startup.

**Verdict: REJECTED.** `gofmt`, `go vet`, and the `execution`/`config`/`daemon`/`runner` suites are all clean. Add `exec` and `mode=1777` to both tmpfs mounts, add a docker test with an image that pre-creates `/home/apiary` and one that execs from `$HOME`, fix the `config.go` comments, and re-request.

orlandoburli:
## Review Council — Round 4 (commit `47c49ff`): **PASS** (5/5 APPROVE)

Both round-3 vetoes are fixed and **independently verified against real Docker**, not taken on the author's word.

### Veto 1 — tmpfs noexec: FIXED (verified)
Raw `docker run` on this machine, same mount strings the code emits:

```
OLD "rw,nosuid,size=512m"                   -> tmpfs (rw,nosuid,nodev,noexec,relatime,size=524288k,mode=755)
NEW "rw,exec,nosuid,mode=1777,size=512m"    -> tmpfs (rw,nosuid,nodev,relatime,size=524288k)
```
`noexec` is gone and `mode=755` is gone. Confirmed for both `/tmp` and `$HOME`.

### Veto 2 — tmpfs inherits mountpoint mode: FIXED (verified)
Built an image pre-creating `/home/apiary` as root:0755 and ran by hand:
- old flags: `touch: /home/apiary/x: Permission denied` -> `DENIED`
- new flags: `WROTE`, and a chmod+x script in `$HOME` printed `EXECOK`

### Regression tests are genuine, not decorative
Reverted **only** the two mount strings in a scratch copy of the tree and re-ran:
```
--- FAIL: TestSandbox_HomeWritableWhenImagePreCreatesIt  err=exit status 1   (Permission denied)
--- FAIL: TestSandbox_TmpfsIsExecutable                  err=exit status 126 (Permission denied)
```
Exactly the failure modes claimed. Restored, all 7 container tests pass (4.9s). This is the standard round 3 asked for: a test that fails against the broken sandbox.

### extra_args allow-list still holds after `--dns*` removal
Re-ran ~65 hostile inputs. Six *appeared* accepted; all six are value-consumption, and I confirmed against real docker that pflag consumes the token identically, so none is a bypass:
- `--label --privileged` -> container ran with `CapEff: 0000000000000000` (vs `000001ffffffffff` when genuinely privileged)
- `-m --privileged`, `--pull --privileged` -> docker **errors out**, container never starts (fail-closed)
- `--hostname -v` -> hostname is literally `-v`; nothing mounted
`-v/:/hostfs`, `-u0:0`, `--read-only=false`, `--cap-add`, `--userns=host`, `--pid=host`, `--device`, `--mount`, `--entrypoint`, `--dns` all rejected. No bypass found.

### Other round-3 items
- **MCP fail-closed ordering**: correct. The check sits at cli.go:86-89, `setupMCP()` at :94-99 — before any side effect. It **does** cover agent-scope MCPs: `runnerConfigWithMCPs` merges runner+agent servers into `config["mcps"]` before `Configure` is called, so the merged set is what's tested.
- `--dns*` removed; `config.go` uid:gid comment corrected; `ExtraArgs` doc example now `["--memory","4g","--pids-limit","512"]` (passes the validator); docs `mode=1777`/`exec` claim now unconditionally true; `HostWiringNotForwarded` sets a real host `SSH_AUTH_SOCK`.
- **Commit message is now accurate**, including the previously-false tmpfs claims. "seven tests execute real containers" — verified, exactly 7. Egress honestly scoped as out of contract ("Refs", not "Closes").
- `gofmt -l` clean, `go vet` clean, all 5 touched packages pass **with and without** `APIARY_SKIP_DOCKER_TESTS=1`.

### Lenses
- **Correctness Hawk — APPROVE.** Both breakages reproduced, fixed, and locked by tests that provably fail on revert.
- **Security Skeptic — APPROVE.** No `extra_args` escape after two rounds of hostile input; DOCKER_HOST/SSH_AUTH_SOCK contained; secrets by name only; MCP fails closed. Egress gap is documented, not hidden.
- **Conventions Stickler — APPROVE (was the holdout).** Config comments, docs, and schema now agree with the code. Only gap left is the missing example yaml — non-blocking for an opt-in feature that ships a complete yaml snippet in docs/runners.md plus a schema entry.
- **Simplicity Advocate — APPROVE.** Allow-list is the simpler and only reasonable-to-audit design; nothing gold-plated.
- **Pragmatist — APPROVE.** Opt-in, no behaviour change for runners without `sandbox`.

### Remaining defects (all non-blocking, follow-up)
1. No sandbox/`env_passthrough` example in `.apiary/example-apiary-full.yaml`.
2. Allow-list contains `--platform`, `--pull`, `--quiet`, `--name`, `--hostname` — benign, but the error string and docs say "resource limits and labelling only", which slightly understates the set.
3. `mode=1777` gives a world-writable sticky HOME. Correct pragmatic choice given the uid is operator-configurable; `uid=<daemon uid>,mode=0700` would be tighter. Single-agent ephemeral container, so impact is negligible.
4. `--oom-kill-disable` is allowed (operator-set, so theoretical host memory-pressure DoS only).
5. `TestSandbox_HostWiringNotForwarded` still lists `DOCKER_HOST` in envNames without asserting on it at container level — already covered at argv level by `TestWrapCommand_DoesNotForwardHostWiringIntoContainer`.
6. Network egress remains uncontained by design; #289 stays open.

PR #296, opened 2026-07-25
security: TLS for the worker queue transport (SEC-13)

orlandoburli:
## Summary

- Adds `tls_cert_file` / `tls_key_file` config fields to `settings.queue`; when both are set the daemon wraps the TCP listener with `crypto/tls` (TLS 1.2+), so the worker bearer token is never sent in plaintext.
- Adds `tls_proxy_mode: true` as an explicit opt-in for deployments where a TLS-terminating proxy (nginx, Caddy, Traefik, etc.) fronts the queue port — gives operators a deliberate escape hatch instead of silently allowing plaintext.
- **Startup gate**: if the listen address is non-loopback (`0.0.0.0`, `::`, or any routable IP) and neither TLS cert/key nor `tls_proxy_mode` is configured, the daemon refuses to start with a clear error message pointing at both remediation paths.
- Config validation rejects cert-without-key (or vice-versa) and the combination of `tls_proxy_mode: true` with cert/key fields.
- New unit tests: `isLoopbackOnly` helper coverage and extended `validateQueueSettings` table.

## Test plan

- [x] `go build ./...` — clean compile
- [x] `go vet ./...` — no issues
- [x] `go test ./internal/config/... ./internal/queuehttp/...` — all pass
- [x] `go test ./internal/daemon/ -run TestIsLoopbackOnly` — passes
- [ ] CI: lint + full test suite

Closes #236

### DIFF
--- a/src/internal/config/config.go
+++ b/src/internal/config/config.go
@@ -306,6 +306,9 @@ type QueueSettings struct {
 	PollInterval       string                   `yaml:"poll_interval,omitempty"`
 	Listen             string                   `yaml:"listen,omitempty"`
 	WorkerToken        string                   `yaml:"worker_token,omitempty"`
+	TLSCertFile        string                   `yaml:"tls_cert_file,omitempty"`
+	TLSKeyFile         string                   `yaml:"tls_key_file,omitempty"`
+	TLSProxyMode       bool                     `yaml:"tls_proxy_mode,omitempty"`
 	Concurrency        QueueConcurrencySettings `yaml:"concurrency,omitempty"`
 }
 
--- a/src/internal/config/queue_test.go
+++ b/src/internal/config/queue_test.go
@@ -14,6 +14,9 @@ func TestValidateQueueSettings(t *testing.T) {
 		{"listener requires token", QueueSettings{Listen: ":8080"}, "worker_token is required"},
 		{"heartbeat before lease", QueueSettings{LeaseDuration: "5s", HeartbeatInterval: "5s"}, "must be shorter"},
 		{"positive scoped limit", QueueSettings{Concurrency: QueueConcurrencySettings{Pools: map[string]int{"build": 0}}}, "positive limit"},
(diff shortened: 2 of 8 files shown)


orlandoburli:
**Reviewer notes — SEC-13 TLS for the worker queue transport**

Reviewed the implementation — the security objective is met.

**What the change does (verified):**

- Non-loopback queue listeners without `tls_cert_file`/`tls_key_file` or `tls_proxy_mode: true` cause a fail-fast startup error with a clear, actionable message.
- `tls.NewListener` wraps the TCP socket when cert+key are provided; `MinVersion: tls.VersionTLS12` is set.
- Config validation rejects cert-without-key (and vice-versa), and `tls_proxy_mode` + cert/key together.
- `isLoopbackOnly` correctly handles IPv4 loopback, `[::1]`, `localhost` (case-insensitive), and returns `false` for wildcards (`::`, `0.0.0.0`).
- `build ./...`, `vet ./...`, and all TLS-specific tests pass (`TestIsLoopbackOnly`, `TestValidateQueueSettings`).

**Minor observations (non-blocking):**

1. `queueControlPlaneURL` always emits `http://` for the embedded worker even when `tls_cert_file`/`tls_key_file` are set. Consequence: configuring direct TLS on a loopback listener would cause the embedded worker to send plaintext to a TLS port and fail. In practice the combination is unusual (loopback traffic is exempt from the plaintext guard, so there's no reason to enable TLS there), but a YAML comment on the config fields or a validator warning would help operators avoid the pitfall.

2. `http.Server` has no `ReadTimeout` or `WriteTimeout` — only `ReadHeaderTimeout: 5s`. Low risk for an internal protocol with bounded JSON bodies, but worth a follow-up if the queue server is ever exposed to untrusted clients.

Neither item is a blocker. The core security requirement — remote worker traffic is encrypted or explicitly gated behind a proxy — is satisfied. ✅

PR #347, opened 2026-07-26
security: privilege ceiling — root warning + opt-in least-privilege agent tools [#290]

orlandoburli:
## Summary
Privilege-ceiling controls for #290, reworked after two council reviews. **Both controls are opt-in and default to today's behaviour — merging this PR does not harden an existing fleet.** It adds the mechanism and makes the risk visible; an operator opts in.

## What it actually does
| Setting | Default | Effect |
|---|---|---|
| `settings.refuse_root` | `false` | Running as root (euid 0) logs a loud **warning**. Set `true` to make it a startup **error**. |
| `settings.least_privilege_agents` | `false` | Set `true` to deny `edit`/`bash`/`webfetch` by default for agents on runners supporting per-agent permissions. |
| `agents[].permissions` | — | Per-agent `allow`/`deny` per tool. **Always wins** over the baseline, in both directions. |

## Why opt-in rather than deny-by-default
The first revision flipped the defaults and was vetoed for good reason: agent files are rewritten on **every dispatch**, so an upgrade would silently strip `edit`/`bash` from every existing agent — and the failure mode is silent (the agent runs, produces no diff, burns `max_attempts`). The hard root refusal likewise broke the documented `apiary service install` path, whose systemd unit runs as root.

Defaulting to warn matches the posture already set for loose `.env` permissions in ff779ef (warn-and-load, not refuse).

## Honest scope limits
- **Merging this hardens nothing by itself.** Both switches are off.
- **Only the OpenCode runner** honours per-agent tool permissions. `claude-cli` and `cursor-cli` are unconstrained by this change.
- **Env allowlisting is not here** — that is SEC-07 (#253). The previous attempt (#267) was vetoed for filtering `os.Environ()` while passing `req.Env` through unfiltered; this PR deliberately does not touch env.

## Other fixes from review
- `schema/apiary.json` gains `agents[].permissions`, `settings.refuse_root`, `settings.least_privilege_agents` — previously `additionalProperties: false` rejected the exact fields the error text told users to add. Surgical +24 insert, no whole-file re-indent.
- `apiary validate` rejects unknown tool names and values other than `allow`/`deny` (`bash: allwo` used to be written through verbatim).
- `permissionMap` no longer emits `webfetch` on the permissive default path: the JSON writer historically omitted it, so unifying upward would have been a net **privilege increase**. Regression-tested.
- Fixed dangling `writeOpencodeAgent` / `validFallbackStrategies` doc comments.
- Documented both settings and `agents[].permissions` in `docs/configuration.md`.

## Tests
Permissive default, least-privilege opt-in, explicit override in both directions, and the no-privilege-increase guard. `gofmt`, `go vet`, and `daemon` + `config` package tests pass.

Refs #290 (deliberately not "Closes": both defaults are off, only one runner is covered, and env allowlisting remains open in #253)

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -182,6 +182,7 @@ agents:
 | `skills` | no | Skill names injected into the agent's context |
 | `max_workers` | no | Per-agent concurrency cap (default 1) — see [concurrency](#concurrency) |
 | `max_turns` | no | Max agent turns per step run (default 0 = unlimited). CLI runners pass it as the provider's turns flag (e.g. claude's `--max-turns`); 0 omits the flag |
+| `permissions` | no | Per-agent tool permissions for runners that support them (OpenCode). Keys: `read`, `glob`, `grep`, `task`, `edit`, `bash`, `webfetch`; values `allow`/`deny`. An explicit entry always wins over `settings.least_privilege_agents`. Note: with the permissive default, `webfetch` is omitted from the generated `opencode.json` (preserving the historical key set) unless you set it explicitly |
 | `source_token` | no | Source API token for this agent's write operations — see [Agent identity](#agent-identity) |
 | `source_email` / `source_name` | no | Git author identity exported to the runner environment |
 | `fallbacks` | no | Ordered `{runner, model}` list for [runner failover](resilience.md#runner-failures--failover); `model` optional (empty = that runner's default) |
@@ -281,6 +282,8 @@ settings:
 | `result_comment` | `false` | Post the agent's final output back as a comment |
 | `task_timeout` | `30m` | Default per-run timeout (e.g. `2h`) |
 | `max_attempts` | 3 | Stop re-dispatching a `(task, workflow)` after this many **consecutive failed** instances; `<=0` disables — see [resilience](resilience.md#re-dispatch-failure-cap) |
+| `refuse_root` | `false` | Make running as root (euid 0) a startup **error** instead of a warning. Agent CLIs inherit the daemon's uid, so a prompt-injected agent would execute as root. Default warns so existing root service installs keep working |
+| `least_privilege_agents` | `false` | Deny `edit`/`bash`/`webfetch` by default for agents on runners that support per-agent permissions (OpenCode). Individual agents override via `agents[].permissions` |
 | `default_fallbacks` | — | Global [fallback chain](resilience.md#fallback-chains) inherited by any agent without explicit `fallbacks[]` |
 | `credit_exhausted_cooldown` | `24h` | How long a runner type is paused after a credit-exhausted failure — see [resilience](resilience.md#runner-failures--failover) |
 | `log_max_size_mb` | 50 | Rotate `apiary.log` past this size (MB); negative disables rotation |
--- a/schema/apiary.json
+++ b/schema/apiary.json
@@ -260,6 +260,20 @@
             "description": "Worker capabilities required before a workflow can be claimed",
             "items": { "type": "string" }
           },
+          "permissions": {
+            "type": "object",
+            "description": "Per-agent tool permissions for runners that support them (OpenCode). An explicit entry always wins over the baseline set by settings.least_privilege_agents",
+            "additionalProperties": false,
+            "properties": {
+              "read": {"type": "string", "enum": ["allow", "deny"]},
+              "glob": {"type": "string", "enum": ["allow", "deny"]},
+              "grep": {"type": "string", "enum": ["allow", "deny"]},
+              "task": {"type": "string", "enum": ["allow", "deny"]},
+              "edit": {"type": "string", "enum": ["allow", "deny"]},
+              "bash": {"type": "string", "enum": ["allow", "deny"]},
+              "webfetch": {"type": "string", "enum": ["allow", "deny"]}
+            }
+          },
           "workspace_affinity": {
             "type": "boolean",
             "description": "Pin retries for this task to the worker that first claims it",
@@ -475,6 +489,16 @@
           "type": "string",
           "description": "Default task timeout (e.g., '2h'). Defaults to 120m",
           "pattern": "^([0-9]+(\\.[0-9]+)?(ns|us|µs|ms|s|m|h))+$"
+        },
+        "refuse_root": {
+          "type": "boolean",
+          "description": "Make running as root (euid 0) a startup error instead of a warning. Agent CLIs inherit the daemon's uid, so a prompt-injected agent would execute as root",
+          "default": false
+        },
+        "least_privilege_agents": {
+          "type": "boolean",
+          "description": "Deny edit/bash/webfetch by default for agents on runners that support per-agent permissions (OpenCode). Individual agents override via agents[].permissions",
+          "default": false
         },
         "max_attempts": {
           "type": "integer",
--- a/src/internal/config/config.go
+++ b/src/internal/config/config.go
@@ -128,6 +128,14 @@ type AgentConfig struct {
 	// WorkspaceAffinity pins retries/resumes to the first worker that claims the
 	// task, preserving a local checkout or other non-portable environment.
 	WorkspaceAffinity bool `yaml:"workspace_affinity,omitempty"`
+	// Permissions overrides the tool permissions written into the OpenCode agent
+	// file. Keys are tool names ("bash", "edit", "webfetch", "read", "glob",
+	// "grep", "task"); values are "allow" or "deny". Keys left out follow the
+	// baseline: permissive by default, or deny for edit/bash/webfetch when
+	// settings.least_privilege_agents is set. An explicit entry always wins, so a
+	// single agent can be restricted without changing the global default —
+	// e.g. permissions: {bash: deny} on a review-only agent.
+	Permissions map[string]string `yaml:"permissions,omitempty"`
 }
 
 // FallbackConfig is one entry in an agent's rate-limit failover chain. Runner
@@ -257,11 +265,23 @@ type TasksConfig struct {
 }
 
 type Settings struct {
-	Concurrency   int    `yaml:"concurrency"`
-	LogLevel      string `yaml:"log_level"`
-	StateLock     bool   `yaml:"state_lock"`
-	ResultComment bool   `yaml:"result_comment"`
-	TaskTimeout   string `yaml:"task_timeout"`
+	Concurrency int    `yaml:"concurrency"`
+	LogLevel    string `yaml:"log_level"`
+	StateLock   bool   `yaml:"state_lock"`
+	// RefuseRoot makes running as root (euid 0) a startup error instead of a
+	// warning. Agent CLIs inherit the daemon's uid, so a prompt-injected agent
+	// executes with whatever privilege the daemon has. Default false — Apiary
+	// warns but starts, so existing root service installs keep working; set true
+	// to enforce a non-root daemon.
+	RefuseRoot bool `yaml:"refuse_root,omitempty"`
+	// LeastPrivilegeAgents makes agent tool permissions deny-by-default
+	// (read/glob/grep/task allowed; edit/bash/webfetch denied) for runners that
+	// support per-agent permissions. Default false preserves the historical
+	// permissive behaviour; individual agents can always restrict themselves via
+	// agents[].permissions regardless of this setting.
+	LeastPrivilegeAgents bool   `yaml:"least_privilege_agents,omitempty"`
+	ResultComment        bool   `yaml:"result_comment"`
+	TaskTimeout          string `yaml:"task_timeout"`
 	// MaxAttempts caps how many consecutive FAILED workflow instances a single
 	// (task, workflow) pair may accumulate before the dispatcher stops
 	// re-dispatching it and applies the workflow's on_fail hook. Rate-limited
@@ -783,6 +803,7 @@ func Load(path string) (*Config, error) {
 		cfg.Settings.Memory.MaxEntryBytes = 16384
 	}
 	warnDeprecatedResultComment(&cfg)
+	warnUnenforceablePermissions(&cfg)
 	return &cfg, nil
 }
 
@@ -798,6 +819,33 @@ func warnDeprecatedResultComment(cfg *Config) {
 	for _, wf := range cfg.Workflows {
 		if wf.ResultComment != "" {
 			aplog.Warn("workflow %q: %s", wf.ID, msg)
+		}
+	}
+}
+
+// warnUnenforceablePermissions warns when per-agent tool permissions cannot be
+// enforced for an agent's runner. Only the OpenCode runner writes an agent file
+// carrying tool permissions; on other runners least_privilege_agents and
+// agents[].permissions are silently inert, which would otherwise leave an
+// operator believing a fleet is restricted when it is not.
+func warnUnenforceablePermissions(cfg *Config) {
+	adapters := map[string]string{}
+	for i := range cfg.Runners {
+		adapters[cfg.Runners[i].ID] = cfg.Runners[i].AdapterName()
+	}
+	for _, a := range cfg.Agents {
+		runnerID := a.Runner
+		if runnerID == "" {
+			runnerID = cfg.DefaultRunner
+		}
+		adapter := adapters[runnerID]
+		if adapter == "" || strings.HasPrefix(adapter, "opencode") {
+			continue // unknown (reported elsewhere) or supported
+		}
+		if len(a.Permissions) > 0 {
+			aplog.Warn("agents[%q]: permissions are ignored by runner %q (adapter %q); only the opencode runner enforces per-agent tool permissions", a.ID, runnerID, adapter)
+		} else if cfg.Settings.LeastPrivilegeAgents {
+			aplog.Warn("settings.least_privilege_agents is set but agent %q uses runner %q (adapter %q), which does not enforce per-agent tool permissions; this agent is NOT restricted", a.ID, runnerID, adapter)
 		}
 	}
 }
--- a/src/internal/config/permissions_warn_test.go
+++ b/src/internal/config/permissions_warn_test.go
@@ -1,0 +1,25 @@
+package config
+
+import "testing"
+
+// least_privilege_agents and agents[].permissions are inert on runners other
+// than opencode; the operator must be told rather than silently believing the
+// fleet is restricted.
+func TestWarnUnenforceablePermissions_DoesNotPanicAndSkipsOpencode(t *testing.T) {
+	cfg := &Config{
+		DefaultRunner: "claude",
+		Runners: []RunnerConfig{
+			{ID: "claude", Type: "cli", Provider: "claude"},
+			{ID: "oc", Type: "cli", Provider: "opencode"},
+		},
+		Agents: []AgentConfig{
+			{ID: "on-claude", Permissions: map[string]string{"bash": "deny"}},
+			{ID: "on-opencode", Runner: "oc", Permissions: map[string]string{"bash": "deny"}},
+			{ID: "unknown-runner", Runner: "nope"},
+		},
+		Settings: Settings{LeastPrivilegeAgents: true},
+	}
+	// Exercises every branch: unsupported adapter with explicit permissions,
+	// supported adapter, and an unresolved runner (reported by Validate instead).
+	warnUnenforceablePermissions(cfg)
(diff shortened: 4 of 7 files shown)


orlandoburli:
## 🐝 Reviewer Council — PR #347 (rework of #267, issue #290)

**VERDICT: REJECTED** — Correctness Hawk veto.

The two controls are the right controls and the env-filtering deferral to SEC-07 (#253, open) is legitimate scoping — this rework does not repeat #267's veto. It is rejected for shipping a **silent, undocumented breaking default flip** plus a set of stale mirrors that make the documented remediation unusable.

---

### 1. Correctness Hawk — **REJECT**

**C1 (blocking) — every existing OpenCode agent silently loses `edit` and `bash` on upgrade, with no migration path shipped.**
`writeOpencodeAgent` / `registerAgentInConfig` are re-run on every dispatch (`dispatcher.go:370`, `:396`, `:1937`), so an upgraded daemon rewrites `.opencode/agents/<id>.md` and the `opencode.json` entry with `edit: deny, bash: deny` for every agent that has no `permissions:` block. That is 100% of existing configs. The failure mode is not a startup error — the agent runs, cannot write files or run git, produces no diff, and burns `max_attempts` retries. Intentional breakage is fine; **undocumented** breakage is not. Nothing in this PR updates:
- `docs/configuration.md` — `agents[].permissions` is not documented anywhere.
- `.apiary/example-apiary-full.yaml` — ships `opencode-cli` agents (lines 203, 210) with no `permissions:`, i.e. the shipped example now yields an engineer agent that cannot edit code out of the box.
- `openspec/CHANGELOG.md` / any release-note migration section.

**C2 (blocking) — `schema/apiary.json` is stale and *rejects the remediation*.**
Both `agents[].items` and `settings` are `"additionalProperties": false`, and neither `permissions` nor `allow_root` was added. A user who follows the error message and writes `settings.allow_root: true`, or who restores their agent with `permissions: {bash: allow}`, gets a red squiggle from the schema this repo ships and documents in `docs/SCHEMA_SETUP.md`. The repo convention is to re-mirror `schema/apiary.json` on config-struct changes; that was skipped.

**C3 — `apiary service install` produces a root daemon, and the docs still say so.**
`src/internal/cli/service.go` builds `kardianos/service` with no `UserName` set, so the generated systemd system unit runs as **root**. `docs/cli.md:184` and `docs/quickstart.md:140` present `apiary service install` as the supported install path. After this PR that path fails to start with the new euid-0 error, and no doc tells the operator to set `User=` or `allow_root`. (Credit where due: the *Docker* path is fine — the image is `distroless/static:nonroot`, so containers are unaffected. The PR body's "opt-in remains for container deployments" is arguing against a problem it doesn't have.)

**C4 — dangling doc comment.** The `// writeOpencodeAgent writes a markdown agent file…` block was left in place and is now glued directly above the `agentPermissions` comment (`dispatcher.go:1715-1721`). `writeOpencodeAgent` has lost its doc comment entirely and `agentPermissions` has a bogus three-line preamble about writing markdown files. Trivially fixable, but it is exactly the kind of thing a rework should not introduce.

### 2. Security Skeptic — **APPROVE** (with recorded gaps)

The root refusal is correct: `os.Geteuid() == 0` is the right predicate (not `Getuid`), it is checked first in `daemon.New` before any work, it covers both the `run` and `worker` entry points, and it returns `-1` on Windows so there is no false positive. The error message names the escape hatch. `settings.allow_root` defaulting to `false` is fail-closed and the right polarity.

Dropping env-filtering is **not** abandoning the control: #253 (SEC-07) is open and covers exactly the merged-env allowlist that #290 asked for. Shipping a correct half beats re-shipping the broken half. No veto.

Recorded gaps for the follow-up, none of them regressions:
- The tool ceiling covers **only** the OpenCode runner. `claude-cli` gets no tool restriction, and `cursor-cli` is spawned with `--force` (`runner/providers/providers.go:57`). An operator who believes "bash is denied" is wrong for 3 of 4 runners, and nothing warns them. The PR body's "this is *the* per-agent privilege ceiling" overstates the coverage.
- `agentPermissions` passes values through unvalidated — `permissions: {bash: allwo}` is written verbatim into `opencode.json` with no lint error. Add a config-lint check that values are `allow|deny|ask`.
- No warning is logged when an agent is granted `bash: allow` / `edit: allow`; a one-line startup log would make the ceiling auditable.

### 3. Conventions Stickler — **REJECT**

Stale mirrors (C2), undocumented public config surface (C1), dangling godoc (C4). Additionally: the PR says `Closes #290`, but #290's stated fix direction is "apply the allowlist to the merged environment; add a test asserting a non-allowlisted key in `req.Env` is dropped." This PR implements none of that. Auto-closing #290 is wrong bookkeeping even though #253 covers the work — either re-scope #290 explicitly or drop the `Closes`.

### 4. Simplicity Advocate — **APPROVE**

`agentPermissions` + `permissionMap` is the minimal correct factoring, and it fixes a real latent bug: the markdown file and the JSON entry previously disagreed (the JSON entry omitted `webfetch` entirely). Driving both from one map is strictly better. The ordered slice in the markdown writer keeps output deterministic — good, since the file is rewritten every dispatch and a map range would have caused spurious churn.

### 5. Pragmatist — **REJECT**

The scoping call is right and I would happily merge the root check today. But this lands on users as: daemon won't start on the documented systemd path, and every agent quietly stops being able to edit code, with no changelog entry and a schema that rejects the fix. That is a support-ticket generator, not a security win. This is ~45 minutes of work away from mergeable.

---

### Tally
| Lens | Verdict |
|---|---|
| Correctness Hawk (veto) | ❌ REJECT |
| Security Skeptic (veto) | ✅ APPROVE |
| Conventions Stickler | ❌ REJECT |
| Simplicity Advocate | ✅ APPROVE |
| Pragmatist | ❌ REJECT |

2/5 approve, and a Correctness Hawk veto → **REJECTED**.

### To land this
1. Add `allow_root` (settings) and `permissions` (agents) to `schema/apiary.json`.
2. Document `agents[].permissions` in `docs/configuration.md` with the default table, and add `permissions: {bash: allow, edit: allow}` to the opencode agents in `.apiary/example-apiary-full.yaml`.
3. Add a **BREAKING CHANGE** entry to `openspec/CHANGELOG.md` / release notes: "OpenCode agents now default to deny for edit/bash/webfetch; add `permissions:` to restore."
4. Document the non-root requirement in `docs/installation.md` + `docs/cli.md`, and set `UserName` in `svcConfig` (or at minimum warn on `service install`).
5. Delete the orphaned `writeOpencodeAgent` comment lines above `agentPermissions` and restore it above the function.
6. Lint `permissions` values to `allow|deny|ask`.
7. Drop or re-scope `Closes #290`.

orlandoburli:
## Council re-review — PR #347 (a857c73): **REJECTED** (2/5 approve; Security Skeptic veto)

The **code** is in good shape this round. What sinks it is the PR body: it was never updated for the rework, it still says `Closes #290`, and it describes the exact behaviour the council vetoed last time.

### Prior findings 1–8

| # | Finding | Status |
|---|---|---|
| 1 | Silent breaking change on upgrade | **Fixed.** Verified `agentPermissions(ac,false)` returns `allow` for all 7 tools. Markdown frontmatter emits the same 7 keys as before (order changed only — YAML mapping, no semantic effect). |
| 2 | `additionalProperties:false` vs. missing schema fields | **Fixed.** Semantic diff of `schema/apiary.json` vs `main` shows *only* additions: `settings.refuse_root` (bool), `settings.least_privilege_agents` (bool), `agents[].permissions` (object, `additionalProperties:false`, 7 tool keys, `enum:[allow,deny]`). File parses; no properties removed. |
| 3 | `apiary service install` broke under root refusal | **Fixed.** Root path warns by default; error only under `settings.refuse_root: true`. The kardianos unit (still no `UserName`) starts. |
| 4 | Example config broken out of the box | **Moot** — default is permissive, `permissions:` is optional. `.apiary/example-apiary-full.yaml` validates clean (only pre-existing missing-soul-file errors). |
| 5 | Dangling doc comment | **Fixed.** |
| 6 | Coverage overstated (OpenCode only) | **Fixed in docs** (`docs/configuration.md` says "runners that support per-agent permissions (OpenCode)") — **but not in the PR body.** |
| 7 | No value validation | **Fixed and verified live.** `apiary validate` on an injected `{bash: allwo, nosuchtool: allow}` emits both:<br>`✗ agents[0] "investigator": permissions.bash: "allwo" must be "allow" or "deny"`<br>`✗ agents[0] "investigator": permissions: unknown tool "nosuchtool"; supported: bash, edit, glob, grep, read, task, webfetch` |
| 8 | `Closes #290` was wrong | **NOT fixed.** The PR body still reads `Closes #290`. |

Build/test: `gofmt -l` clean on all four touched Go files; `go vet` clean; `go test ./internal/daemon/ ./internal/config/` both `ok`.

### Blocking defects

**B1 — The PR description is stale and now describes the opposite of the code.** It still states: *"`daemon.New` returns an error if `os.Geteuid() == 0` unless `settings.allow_root: true`"* and *"OpenCode agent files … **now default** to `read/glob/grep/task: allow` and `edit/bash/webfetch: deny`."* Neither is true of a857c73: root only warns, defaults are fully permissive, and `allow_root` does not exist anywhere in the tree (it is `refuse_root`, inverted). The squash-merge commit message is the PR body — main's history would permanently claim least-privilege is on by default when it is off. Anyone reading it would believe the daemon is hardened when it is not.

**B2 — `Closes #290` on a control that ships OFF.** This auto-closes the SEC privilege-ceiling issue while: env allowlisting is deferred to #253, only the OpenCode runner is constrained (claude-cli/cursor-cli untouched), and *both* controls default to the pre-PR behaviour. Closing the tracking issue is how the residual work disappears. Use `Refs #290` and leave #290 open with the remaining scope enumerated. This is the difference between "opt-in hardening, honestly scoped" and security theatre.

### Non-blocking

- **N1 (behaviour change, unflagged).** The old `registerAgentInConfig` JSON permission map had **six** keys — no `webfetch`. `permissionMap(agentPermissions(...))` now writes all seven, so under the permissive default `opencode.json` gains an explicit `webfetch: "allow"` where OpenCode previously applied its own default. In a *privilege-ceiling* PR the one net change to the default path is a privilege **increase**. Either drop `webfetch` from the JSON path under the permissive baseline, or state deliberately that OpenCode's own default was already allow.
- **N2.** `agents[].permissions` is absent from the agent field table in `docs/configuration.md` (~line 175). It is documented only in passing from the `least_privilege_agents` settings row — yet it is the mechanism operators need.
- **N3.** `schema/apiary.json` is +349/−81 for ~28 semantically new leaves; the rest is a whole-file re-indent. Verified harmless via a semantic key-diff, but bundling a reformat into a security PR makes the next reviewer do that work too. Split it.
- **N4.** `permissionMap` is a `map[string]string`→`map[string]any` copy used once; inlining it removes a named function from the API surface.

### On the design question: is opt-in right for a security control?

Honestly: **yes here, no in general.** The specific reason opt-in is correct in this PR is the one the code comment names — these files are rewritten on *every dispatch*, so flipping the default is not a config change the operator opts into at upgrade time, it retroactively rewrites live agents, and the failure mode is silent (agent runs, produces no diff, burns `max_attempts`). A security control whose failure mode is invisible must not default on mid-stream. That reasoning is sound and the code comment/test that encode it are the best part of the diff.

But opt-in only escapes the "theatre" charge if the PR is honest that the fleet is *not* hardened by merging it. Right now the PR text claims the opposite of the code and closes the issue. The correct end-state is a follow-up that flips the default at a major version with a migration note — file it, and reference it from #290.

### Lenses

- **Correctness Hawk — APPROVE.** Default path traced through both writers; identical to pre-PR modulo the `webfetch` JSON key (N1). Tests pin the invariant. vet/gofmt/tests clean.
- **Security Skeptic — REJECT (veto).** Not for the opt-in default, which is justified — for B2. A privilege-ceiling issue cannot be auto-closed by a PR that leaves both controls off, covers one of three runners, and defers env allowlisting; and B1 leaves a permanent history entry asserting hardening that does not exist. N1 is a net privilege increase on the default path in a hardening PR.
- **Conventions Stickler — REJECT.** B1/B2 (stale body, wrong issue verb), N2 (undocumented agent field), N3 (reformat bundled into a security change).
- **Simplicity Advocate — APPROVE.** `agentPermissions` is a clean seed-then-override; the fixed `agentTools` ordering kills file churn. N4 is a nit.
- **Pragmatist — APPROVE.** Substantively responsive to all eight findings bar #8; the blockers are a description edit and a one-word change.

**Verdict: REJECTED.** Fix B1 (rewrite the body to describe opt-in defaults and OpenCode-only scope) and B2 (`Refs #290`, keep the issue open), address N1, and this passes on the next pass — the Go code needs no further changes.

orlandoburli:
## Council re-review — round 3 (ff784f1): **REJECTED** (1 blocking, easily fixed)

The code is good. The *merge metadata* still isn't, and it fails in exactly the way round 2 blocked on.

### Verified fixed
- **N1 (privilege increase) — fixed, and fixed correctly.** `permissionMap` skips `webfetch` when `!leastPrivilege` and the agent has no explicit entry, so `opencode.json` keeps its historical six-key set. Least-privilege writes `webfetch: deny`; an explicit entry is honoured in both directions. `TestPermissionMap_NoPrivilegeIncreaseOnDefaultPath` covers all three. The markdown writer still emits all seven (unchanged from `main`, which already wrote `webfetch: allow`), so the md/json divergence is preserved rather than introduced — correct call.
- **N3 — fixed.** `git diff origin/main -- schema/apiary.json` is now **+24/-0**, two surgical inserts. JSON parses.
- **N2 — fixed.** `docs/configuration.md:185` documents `agents[].permissions` with the tool list and override semantics; both settings rows are accurate.
- **PR body vs code:** I checked every row and every claim. The settings/defaults table matches `Settings.RefuseRoot` / `Settings.LeastPrivilegeAgents` (both `omitempty` bools, zero value = today's behaviour) and `agentPermissions`, where the explicit map is applied last and therefore always wins. The scope-limit section is accurate: only `writeOpencodeAgent`/`registerAgentInConfig` consume permissions, nothing touches env. No overstated claim found.
- `gofmt -l` clean on all four touched files; `go vet` clean; `go test ./internal/daemon/ ./internal/config/` pass.

### B1/B2 (blocking) — the fix landed in the PR body, but this repo doesn't merge the PR body

```
$ gh api repos/orlandoburli/apiary --jq .squash_merge_commit_message
COMMIT_MESSAGES
```

A plain `gh pr merge 347 --squash` (this repo's convention — auto-merge is disabled) composes the squash body from the **commit messages**, not the PR description. Commit `4bc671e` is still on the branch and still says:

> Refuse to start when the daemon runs as root (euid 0) unless `settings.allow_root` is set … OpenCode agent files are now written with least-privilege tool permissions … **denied by default**.
>
> **Closes #290**

So on merge, `main`'s history gets the vetoed hard-refusal / deny-by-default description of a setting (`allow_root`) that does not exist in the code, and **#290 auto-closes** — both defaults off, one runner covered, #253 open. That is B1 and B2 verbatim, still armed. Rewriting the PR body did not disarm them; it only fixed the page nobody merges.

**Fix (either one, then re-ping):**
- `git rebase -i origin/main` and reword `4bc671e` (drop `Closes`, drop `allow_root`/deny-by-default), or
- squash the three commits into one whose message is the current PR body.

Merging with an explicit `--body` is not an acceptable substitute — it relies on the merger remembering, and the default path is the one that fires.

### Non-blocking
- **N4 (still awkward).** `permissionMap(perms, explicit, leastPrivilege)` takes the derived map *and* both of its inputs; the callsite re-computes `agentPermissions(ac, lp)` and passes `ac.Permissions, lp` alongside it. Two of three arguments are redundant. `permissionMap(ac, lp)` calling `agentPermissions` internally would be a strictly smaller surface. Not worth another round on its own.
- The md-vs-json `webfetch` divergence is now deliberate and well documented **in code and in the PR body**, but not in `docs/configuration.md` or the schema description — an operator reading the docs will expect `opencode.json` to carry `webfetch` and it won't. One clause on the `permissions` row would close it.
- No new defect introduced by this round's changes.

### Lenses
- **Correctness Hawk — REJECT (veto).** Go code is correct; the squash message that will actually land on `main` contradicts it.
- **Security Skeptic — REJECT (veto).** A false hardening claim in the history of a security repo, plus auto-closing the privilege-ceiling issue while the ceiling is off by default and only covers one runner, is the security defect here. N1 handled well.
- **Conventions Stickler — APPROVE.** gofmt/vet/tests clean, surgical schema, docs row present, `Refs` not `Closes` in the body.
- **Simplicity Advocate — APPROVE.** Modulo N4.
- **Pragmatist — APPROVE.** One `git commit --amend`-scale change from mergeable.

**2 rejects, both veto-holders → REJECTED.** Reword the commits and this passes.

orlandoburli:
## Council re-review — round 4 @ `405637f`

**Verdict: PASS** (5/5 APPROVE; no Correctness/Security veto)

### The round-3 blocker is resolved
`git log origin/main..405637f` = **exactly one commit**. Repo squash settings are `squash_merge_commit_message: COMMIT_MESSAGES` / title `COMMIT_OR_PR_TITLE`, so the message that lands on main is this commit's message verbatim. Every claim in it was checked against the diff:

- "BOTH DEFAULT OFF" — `RefuseRoot`/`LeastPrivilegeAgents` are plain bools with no defaulting code; zero value = off. ✔
- `settings.refuse_root` warn-vs-error at `os.Geteuid() == 0` in `Dispatcher.New`. ✔
- `settings.least_privilege_agents` denies edit/bash/webfetch, keeps read/glob/grep/task. ✔
- `agents[].permissions` "always wins in both directions" — applied last in `agentPermissions`, and covered by a test that asserts both an explicit deny under the permissive baseline and an explicit allow under least-privilege. ✔
- Scope limits stated plainly (OpenCode only; env allow-listing stays #253) — matches the diff, which touches only the two OpenCode writers. ✔
- Ends in **"Refs #290"**, not "Closes", with the reason spelled out. ✔

No trace of the stale 4bc671e text (`settings.allow_root`, deny-by-default, `Closes #290`) remains.

### Behaviour unchanged by the rewrite
Both writers now go through `agentPermissions`:
- `writeOpencodeAgent` (markdown) emits all seven keys, all `allow` by default — byte-identical set to the old hardcoded block (order changed only, cosmetic in YAML front-matter).
- `registerAgentInConfig` (opencode.json) via `permissionMap` omits `webfetch` on the permissive path, preserving the historical six-key set — no net privilege increase; writes it under least-privilege; honours an explicit entry either way.

### Round-3 nits
1. `permissionMap(ac config.AgentConfig, leastPrivilege bool)` — down to two args, redundant third gone. ✔
2. The md-vs-opencode.json `webfetch` divergence is now documented in the `agents[].permissions` row of docs/configuration.md, and in a doc comment plus a named test (`TestPermissionMap_NoPrivilegeIncreaseOnDefaultPath`). ✔

### Gates
`gofmt -l` on all four touched Go files: clean. `go vet ./internal/daemon/ ./internal/config/`: clean. `go test ./internal/daemon/ ./internal/config/`: both ok. `schema/apiary.json` parses; +24 lines adding `agents[].permissions` (enum-constrained, `additionalProperties:false`) and both settings — which is what unblocks the fields the old error text told users to add. `apiary validate` rejects unknown tool names and non-allow/deny values.

### Lenses
- **Correctness Hawk — APPROVE.** No new defect from the squash; the simplification is a pure signature narrowing with tests pinning all four paths.
- **Security Skeptic — APPROVE.** Opt-in only, so merging changes nothing at runtime — and the commit message now says exactly that, so no one reads this as a fleet hardening. No privilege increase anywhere.
- **Conventions Stickler — APPROVE.** Schema/docs/struct stay mirrored; "Refs" over "Closes" is the honest call given #253 is open.
- **Simplicity Advocate — APPROVE.** One shared baseline function, one fixed ordered tool list, no config-time defaulting to reason about.
- **Pragmatist — APPROVE.** Ships the mechanism plus the knob; the divergence is documented rather than papered over.

### Remaining (non-blocking, follow-up)
- `claude-cli` and `cursor-cli` silently ignore `agents[].permissions` — `least_privilege_agents: true` will read as fleet-wide to an operator whose agents are on those runners. Worth a validate-time warning when the setting is on and a non-OpenCode runner is configured.
- The permissive-path `webfetch` omission is a deliberate wart; fold it away when a future change can afford to normalise both writers to seven keys.
- SEC-07 env allow-listing (#253) is still the larger half of the privilege ceiling.

PR #269, opened 2026-07-24
fix(security): warn on group/world-readable .env file

orlandoburli:
## Summary

- Adds a permission check to both `.env` loading paths (`internal/config` and `internal/cli`) that warns when the file is readable by group or world (mode `& 0o044 != 0`)
- Warning includes a `chmod 0600 <path>` remediation command, mirroring the existing plugin-directory permission stance
- Change is **additive** — the file is still loaded, so no existing setups break
- Unix-specific check lives behind `//go:build !windows`; a no-op stub covers Windows
- New test `TestWarnDotEnvPerms` covers the 0600 (silent), 0644 (warns), and missing-file cases

## Test plan

- [x] `go test ./internal/config/... -run TestWarnDotEnvPerms` — passes, warning emitted on 0644
- [x] `go test ./...` — all packages green
- [x] `go build ./...` — clean build

Closes #248

PR #349, opened 2026-07-26
security: warn (not refuse) on loosely-permissioned .env (rework #269) [#293]

orlandoburli:
Reworks council-rejected #269 (issue #293). #269 did an early `return` that **refused to load** a group/world-readable `.env` while claiming to only warn — a silent startup-breaking regression.

`loadDotEnv` now **warns loudly** (with a `chmod 600` remediation hint) when the `.env` is group/world-accessible, but **always loads it** — matching the warn-and-load intent and never breaking startup for existing default-perm setups.

Tests: a 0644 .env still loads (value reaches env); a 0600 .env loads. `go build` + config tests pass.

Closes #293

### DIFF
--- a/src/internal/config/config.go
+++ b/src/internal/config/config.go
@@ -824,6 +824,15 @@ func loadDotEnv(configPath string) {
 	if err != nil {
 		return
 	}
+	// The .env commonly holds real secrets (e.g. a Jira API key). Warn — but do
+	// NOT refuse to load — if it is group- or world-accessible, so other local
+	// accounts could read those secrets. Warn-and-load avoids breaking startup for
+	// existing setups while still surfacing the risk (SEC #293).
+	if info, statErr := os.Stat(envPath); statErr == nil {
+		if perm := info.Mode().Perm(); perm&0o077 != 0 {
+			aplog.Warn("%s is group/world-accessible (mode %04o) and may contain secrets; run: chmod 600 %q", envPath, perm, envPath)
+		}
+	}
 	for _, line := range strings.Split(string(data), "\n") {
 		line = strings.TrimSpace(line)
 		if line == "" || strings.HasPrefix(line, "#") {
--- a/src/internal/config/dotenv_perms_test.go
+++ b/src/internal/config/dotenv_perms_test.go
@@ -1,0 +1,36 @@
+package config
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+// loadDotEnv must always load the file (warn-and-load), regardless of perms —
+// the value must reach the environment even when the file is loosely permissioned.
+func TestLoadDotEnv_LoadsRegardlessOfPerms(t *testing.T) {
+	dir := t.TempDir()
+	envPath := filepath.Join(dir, ".env")
+	if err := os.WriteFile(envPath, []byte("SEC293_TOKEN=abc123\n"), 0o644); err != nil {
+		t.Fatal(err)
+	}
+	// A group/world-readable .env (0644) must still be loaded, not refused.
+	t.Setenv("SEC293_TOKEN", "")
+	loadDotEnv(filepath.Join(dir, "apiary.yaml"))
+	if got := os.Getenv("SEC293_TOKEN"); got != "abc123" {
+		t.Fatalf("warn-and-load: expected value to be loaded even with 0644 perms, got %q", got)
+	}
+}
+
+func TestLoadDotEnv_LoadsWith0600(t *testing.T) {
+	dir := t.TempDir()
+	envPath := filepath.Join(dir, ".env")
+	if err := os.WriteFile(envPath, []byte("SEC293_SECURE=xyz\n"), 0o600); err != nil {
+		t.Fatal(err)
+	}
+	t.Setenv("SEC293_SECURE", "")
+	loadDotEnv(filepath.Join(dir, "apiary.yaml"))
+	if got := os.Getenv("SEC293_SECURE"); got != "xyz" {
+		t.Fatalf("expected 0600 .env to load, got %q", got)
+	}
+}


orlandoburli:
## 🐝 Apiary Review Council — PR #349

**Verdict: ❌ REJECTED** (Correctness Hawk veto)

Reviewed at head `f00b35a`. `go build ./...` and `go test ./internal/config/` both pass on the PR branch.

### What this rework gets right

The #269 regression is genuinely fixed. `loadDotEnv` has exactly one early `return` (the pre-existing `os.ReadFile` error path, which is correct and unchanged), and the new permission block is a pure side-effect `if` — **no path skips the load**. The mask `0o077` is the right one for a secrets file: it catches group *and* other, read *and* write *and* execute (stricter than the `0o022` write-only mask used in `plugin/discovery_unix.go`, appropriately so). The warning logs only `envPath` and the octal mode — **no secret values leak into the log**, which matters given a live Jira API key is in this file today. `os.Stat` failure is handled (`statErr == nil` guard) and correctly does not block loading.

---

### 1. Correctness Hawk — ❌ REJECT (veto)

The load-always claim holds on Unix. But the *warning* is factually wrong on a supported, shipped platform.

`.goreleaser.yaml:21` builds `goos: [darwin, linux, windows]`. On Windows, Go synthesizes `FileMode.Perm()` as `0666` for any writable regular file (`0444` if read-only) — it does not reflect ACLs. So `perm&0o077` is `0o066`, non-zero, **always**:

- Every Windows user gets this WARN on **every single startup**, unconditionally, forever.
- It asserts something false: the file is not "group/world-accessible" — that is unknowable from `Mode().Perm()` on Windows.
- The remediation is unrunnable: `chmod 600` is not a Windows command. There is no action the user can take to silence it.

A security warning that fires 100% of the time on a platform, with a remediation that cannot be executed, is pure alarm fatigue — it actively degrades the signal this PR exists to create. This is the same class of "the message does not match what the code does" defect that sank #269, just moved from behavior to output.

The repo already solves this exact problem one package over: `plugin/discovery_unix.go` / `plugin/discovery_windows.go`, where `checkDirOwnerOnly` is a documented no-op behind `//go:build windows`. Extract a `checkDotEnvPerms(path)` helper with the same split and the veto lifts.

### 2. Security Skeptic — ✅ APPROVE

Warn-only is the defensible posture *for this PR*. Refusing to load is a self-inflicted availability failure that fixes nothing — the secret is already on disk and already exposed; refusing to read it does not un-expose it, it just takes the daemon down. Detection-then-remediation is right, and the message is actionable on Unix (names the file, the actual mode, and the exact fix).

Two things I want on record as follow-ups, not blockers:

- **No strict opt-in.** A `settings.strict_env_perms: true` that hard-fails would let a security-conscious operator enforce what this PR can only suggest. Warn-only means the warning gets ignored forever and nothing ever changes.
- **Coverage is partial.** There is a *second, untouched* `.env` loader: `cli/root.go:109` runs in `PersistentPreRun` before `config.Load` and calls `godotenv.Load(path)` with **no permission check at all**. Worse, it honors `--env-file`, so a `.env` at an arbitrary path is never checked by anything. Because that loader runs first and neither loader overrides already-set vars, the secrets actually in use may come entirely from the unchecked path. The default case is still covered by the config-dir check, so this is not a veto — but the PR claims more coverage than it has.

### 3. Conventions Stickler — ❌ REJECT

Same root cause as the Correctness veto. `internal/plugin` establishes the in-repo convention for permission checks that cannot be expressed on Windows: a `_unix.go` / `_windows.go` build-tagged helper pair. This PR inlines a raw `Mode().Perm()` check into shared code and ignores that precedent. Notably, a later commit on a sibling branch (`ff779ef`) implements precisely the platform-split `checkDotEnvPerms` helper — the correct shape is already known, it just is not what was pushed here.

### 4. Simplicity Advocate — ✅ APPROVE

Nine lines, no new types, no config surface, no control-flow change. The comment explains *why* warn-and-load rather than restating *what*. Extracting the helper for the Windows fix adds one small file and keeps this just as simple.

### 5. Pragmatist — ✅ APPROVE

Right scope for the issue, zero migration cost, cannot break an existing deployment on Unix. The Windows fix is a ~10-line mechanical change against an existing template — cheap enough that shipping without it is not worth it.

---

### Decision

Correctness Hawk vetoed → **REJECTED** (2 REJECT / 3 APPROVE is moot under the veto rule).

### Required to pass

1. **Blocking** — Extract `checkDotEnvPerms(envPath string)` into `config/dotenv_perms_unix.go` (`//go:build !windows`) and `config/dotenv_perms_windows.go` (`//go:build windows`, documented no-op). No spurious warning on Windows, no `chmod` advice on a platform without `chmod`.

### Strongly recommended

2. Assert the **warning itself** in a test. Both current tests would still pass if the entire warn block were deleted — they prove *load*, not *warn-and-load*. Capture `aplog` output (or route it through an injectable writer) and assert the 0644 case warns and the 0600 case does not.
3. Apply the same check to `cli/root.go:109` `loadDotEnv`, including the `--env-file` path, or explicitly document the gap.
4. Open a follow-up for a `settings.strict_env_perms` opt-in so operators can enforce rather than merely be reminded.

### Nit (non-blocking)

The tests lean on `t.Setenv(key, "")` interacting with `loadDotEnv`'s `if os.Getenv(key) == ""` guard — set-to-empty is indistinguishable from unset, so the load proceeds. Correct today and `t.Setenv` cleanup prevents cross-test pollution (distinct keys, no `t.Parallel()`), but it is an implicit coupling: if the loader ever distinguishes set-but-empty from unset, these tests break for reasons unrelated to what they test. Prefer `os.Unsetenv` inside a `t.Cleanup`.

orlandoburli:
Closing as **superseded by ff779ef** (branch `fix/293-dotenv-warn-and-load`), which fixes #293 better than this PR:

- Uses the established platform split (`//go:build !windows` + `checkDotEnvPerms`), avoiding the false permanent warning this PR would emit on Windows — where Go synthesizes `Mode().Perm()` as 0666 regardless of ACLs, making the check always fire with a `chmod` remediation that doesn't exist on the platform (the council's blocking finding here).
- Stricter, more accurate mask (`0o044`).
- Also covers the second `.env` loader at `internal/cli/root.go` (godotenv, honors `--env-file`), which this PR left unchecked.

No further work needed on #293 beyond landing ff779ef.

Issue #292, opened 2026-07-25
[security] Rework SEC-10: feature-detect userns + drop out-of-scope doc reverts (PR #255 rejected)


orlandoburli:
Council **rejected** PR #255 (Correctness veto). The checksum + path-validation code is good and could ship on its own; the sandbox path is broken.

**Defect:** the default `network:false` path unconditionally requires unprivileged user namespaces with no feature-detection/fallback, so plugins hard-fail on hardened kernels and many container runtimes (including Apiary's own Docker image). Also bundles 6 unrelated gitnexus SKILL.md doc reverts.

**Fix direction:** feature-detect userns and fall back gracefully (warn + run without netns, or refuse only that plugin); remove the out-of-scope doc reverts; reframe "signed" as integrity (digest beside binary), not authenticity.
Ref PR #255 (SEC-10).

orlandoburli:
## Status: split — integrity shipped, network enforcement deferred

PR #350 now covers only **executable integrity pinning** (`checksum`), which passed council review. **Network-isolation enforcement was removed from it** and remains open here.

### Why enforcement was pulled
Council review found it was a breaking change: `SecurityRequirements.Network` is a bare `bool` that was **purely declaratory** before. Enforcing it means a manifest that merely *omits* the field is read as `false` → "isolate", which would silently kill outbound network for every already-installed plugin on Linux — and loopback too, since a fresh netns brings up `lo` down, breaking any plugin talking to 127.0.0.1.

### Requirements for whoever implements this
1. **Tri-state default.** `Network` must become `*bool` (or equivalent) so *absent* = legacy-permissive, distinct from an explicit `false`. Do not silently re-interpret existing manifests.
2. **Bring up loopback** inside the netns, or plugins using 127.0.0.1 break.
3. **Feature-detect, never hard-require.** Probe for a usable unprivileged netns (`unshare -r -n`) with a bounded timeout; hardened kernels, macOS, and container runtimes (including Apiary's own image) can't provide one. PR #255 was rejected for requiring it unconditionally.
4. **Don't cache a failed probe forever.** A transient boot-time failure behind `sync.Once` becomes a permanent silent fail-open.
5. **Strict/enforce mode.** Without an opt-in that makes unenforceable isolation a hard error, `network: false` still can't be *relied* on — which is the substance of the original finding.
6. **`unshare -r`** maps uid 0 *inside* the namespace; document so authors don't read it as host root.
[...]

PR #350, opened 2026-07-26
security: plugin executable integrity (checksum pinning) [#292]

orlandoburli:
## Summary
Scoped rework for #292. Adds **executable integrity pinning** for plugins. Network-isolation enforcement has been **removed from this PR** — see below.

## Why network enforcement was dropped
Council review found a breaking change: `SecurityRequirements.Network` is a bare `bool` that was **purely declaratory** before this work, so a manifest that merely *omits* it would now mean "isolate" — silently killing outbound network, and even loopback (a fresh netns has `lo` down), for every already-installed plugin on Linux.

Fixing that properly needs a tri-state/legacy-permissive default, `lo` setup, a strict-mode opt-in, and a compatibility story. That's a real design change protecting a feature nobody currently runs — the operator-context security review classes plugins as **not applicable** and explicitly **not a rollout gate**. So enforcement is deferred to a dedicated change rather than rushed in here; #292 stays open for it.

## What this PR does
- Manifest may pin `checksum` (`sha256:<hex>` or bare hex), verified **at load and before every invocation**, so a binary replaced while the daemon runs is caught (not boot-only).
- Normalization trims/lowercases **before** stripping the prefix, so `SHA256:<hex>` and padded values work (this was a bug in the first attempt).
- A non-blank but **malformed** pin is an error rather than silently meaning "unpinned", and `Installed.Validate` checks it, so `apiary validate` catches it.
- `integrityGuard` re-hashes only when size/mtime changed — one `stat()` in steady state.
- `docs/plugins.md` documents the field.

## Honest limits (documented in code and docs)
This is **tamper-evidence, not authenticity**: the digest lives beside the binary, so anyone who can rewrite the executable can rewrite the pin. An mtime+size-preserving swap defeats the cache, and there is an unavoidable TOCTOU window before exec. It targets accidental drift and unsophisticated swaps.

## Known follow-up
`checksum` is a new manifest field while `ManifestSchemaVersion` stays 1, and `Load` uses `DisallowUnknownFields`, so an **older** Apiary reading a **newer** manifest fails with `unknown field "checksum"` instead of the friendly version-mismatch error. It fails closed, and bumping to 2 would reject every existing manifest, so it is left as-is deliberately.

## Tests
Normalization table (accepted/unpinned/malformed), tamper detection, guard re-verification after a binary swap. `gofmt`, `go vet`, and the full `plugin` package tests pass.

Refs #292


orlandoburli:
## Council verdict: REJECTED (2 vetoes)

The #255 veto **is** fixed: `applyNetworkIsolation` never hard-fails, feature detection actually executes a probe rather than assuming, and the unrelated SKILL.md reverts are gone. Credit where due. But this rework introduces its own set of concrete defects, several demonstrable.

### Verified findings

**1. Checksum normalization order is wrong (demonstrated).**
```go
want = strings.ToLower(strings.TrimSpace(strings.TrimPrefix(want, "sha256:")))
```
`TrimPrefix` runs *before* `ToLower`/`TrimSpace`, so `SHA256:<hex>` and `" sha256:<hex>"` are **not** stripped. Added a throwaway test on this branch:
```
uppercase prefix rejected: ... manifest pins sha256:sha256:2d7116... but file is sha256:2d7116...
leading space  rejected: ... (same doubled prefix)
```
Fail-closed, but the error message is self-contradictory ("pins sha256:sha256:X but file is X") and a legitimate manifest is bricked. Correct order: TrimSpace → ToLower → TrimPrefix (and re-TrimSpace).

**2. `"checksum": "sha256:"` silently disables the check.** After the prefix strip the value is empty, which is the "unpinned" sentinel. A truncated/garbage-but-empty-after-prefix pin degrades to no verification with no warning. There is also no `len==64 && ishex` validation anywhere, and `Installed.Validate` never inspects `Checksum`, so `apiary validate` cannot catch a malformed pin — it only explodes at daemon start.

**3. Verification is boot-only, which does not match the stated threat model.** `NewClient` is called exactly once, from `EnabledClients` → `NewDispatcher`. The PR claims it "detects post-install tampering/swap-out"; in practice it detects tampering that happened *before the daemon started*. Swapping the binary while the daemon runs — the actual persistence technique — is never detected, for the entire (multi-day) lifetime of the process. Either verify per-invoke (cheap fast path: stat size+mtime, re-hash on change) or stop claiming it catches post-install swap-out.

**4. The probe is unbounded and runs on the invoke hot path.** `detectNetIsolation` runs `exec.Command(unshare, "-r", "-n", "--", "true").Run()` with **no context and no timeout**, lazily, inside `sync.Once`, first triggered from `Invoke`. If `unshare` hangs (odd seccomp/container behaviour is exactly the environment class this PR exists to handle), every plugin invocation blocks on the `Once` forever. Probe at daemon start, with a `CommandContext` + a few-second timeout.

**5. The probe result is cached forever, and both stale directions degrade badly.** Cached "supported" + sysctl later flipped off ⇒ `unshare` fails at exec and surfaces as `plugin %q crashed or exited unsuccessfully` — the operator gets a plugin-crash message for a sandbox problem. Cached "unsupported" never re-evaluates. Also the probe's failure reason is swallowed entirely: no `Debug` log distinguishes "unshare not installed" from "userns disabled by sysctl" from "blocked by seccomp", which is precisely the diagnostic an operator on a hardened host needs.

**6. The enforcement branch is completely untested.** `TestApplyNetworkIsolation_NeverHardFails` accepts *either* branch, so on macOS (dev + likely CI) it only ever exercises pass-through. Nothing tests that `Invoke` still round-trips JSON over stdin through the wrapper. Reading it, the wiring *is* correct — `unshare` without `--fork/--pid` `execvp`s in place, so fd 0/1/2, `Cmd.Dir` and `Cmd.Env` all survive, and `CommandContext`'s kill still lands on the right PID — but "correct by reading" is what sank #255. Needs a Linux-gated integration test.

**7. `unshare -r` is not a neutral wrapper.** `-r` maps the caller to uid 0 in a new user namespace: the plugin now believes it is root and holds full capabilities *within* that namespace (can create further namespaces, mount tmpfs/proc, etc.). That is a real behavioural and security-posture change bundled silently into "deny network". At minimum document it; consider `--map-user=$(id -u)` semantics or dropping `-r` where the kernel allows netns without it.

**8. Manifest compat: new field, no `schema_version` bump.** `Load` decodes with `DisallowUnknownFields()` **before** the `apiary` constraint is checked, so an older Apiary reading a new pinned manifest dies with `unknown field "checksum"` instead of the intended "requires Apiary >= X". The manifest's own version-negotiation mechanism is bypassed by decode order.

**9. gofmt violation.** `gofmt -l internal/plugin/` flags `manifest.go` — the new `Checksum`/`Capabilities` lines were realigned but the rest of the struct was not.

**10. Docs untouched.** `docs/plugins.md` still lists the manifest fields without `checksum`, still says "verify its checksum and signature out of band", and its "Process isolation is not a security sandbox" section is now half-true. No JSON-schema entry either.

### The policy question: is warn-and-continue acceptable?
Yes in principle — silently pretending or hard-failing are both worse, and this is never weaker than today's zero enforcement. But as shipped, an operator who *needs* `network:false` to hold still has no way to demand it. There is no `strict`/`require_isolation` setting that refuses to load an unenforceable plugin, and the warning goes to stderr **on every single invoke** (event-exporter = one WARN line per event → flood, then ignored), while never surfacing in `apiary validate` or `apiary plugins`. That inverts the right cost model: the expensive check (hashing) runs once, the free one (warning) runs thousands of times. Warn once per plugin at startup, surface the unenforced state in `validate`/`plugins` output, and add an opt-in strict mode.

---

### Lenses

- **Correctness Hawk — REJECT (veto).** Findings 1, 2, 4, 5: a demonstrated normalization bug that rejects valid `SHA256:` pins with a self-contradictory message, a silent-disable sentinel, and an untimed subprocess probe serialized behind `sync.Once` on the invoke path that can wedge every plugin call.
- **Security Skeptic — REJECT (veto).** Finding 3: verification happens once at daemon boot, so the advertised "detects post-install tampering/swap-out" does not hold against the runtime-swap threat it names. Compounded by finding 7 (`unshare -r` hands the plugin uid 0 + full caps in a new userns as an undocumented side effect) and the absence of any strict mode.
- **Conventions Stickler — REJECT.** gofmt-dirty file (9), new manifest field with no `schema_version` bump and no schema/doc update (8, 10), probe failure swallowed with no `Debug` log contrary to the repo's fail-loud norm.
- **Simplicity Advocate — APPROVE.** ~110 lines, two orthogonal mechanisms, no config surface, no new deps; the argv-prefix shape is the simplest thing that could work.
- **Pragmatist — APPROVE.** Strictly better than the status quo, unblocks #292's direction, and every finding above is a one-pass fix. Land it after the fixes rather than redesigning.

**2 vetoes → REJECTED.** Required before re-review: fix (1) and (2) with tests, timeout + startup-time probe for (4), per-invoke or mtime-gated re-verification (or retract the claim) for (3), Linux-gated wrapper integration test (6), `gofmt` (9), and manifest/docs/schema updates (8, 10). Findings 5, 7 and the warn-once/strict-mode point should land in the same pass.

orlandoburli:
## Council re-review of 5002cc8 — REJECTED (2/5)

Verified locally: `gofmt -l internal/plugin/` clean, `go vet` clean, `go test ./internal/plugin/` **ok** (0.98s). Findings 1, 2, 3, 4, 5, 7 and 9 from the previous round are genuinely fixed and I could not break them:

- `normalizeChecksum` now trims+lowercases **before** `TrimPrefix`; `SHA256:<HEX>`, ` sha256:<hex> ` and bare hex all normalize correctly, and `"sha256:"`, `"deadbeef"`, non-hex are hard errors rather than silent "unpinned". `Validate` calls it, so `apiary validate` surfaces a bad pin. Good.
- `integrityGuard.check` really does re-hash after a swap (test proves it); mutex is held for the compare-and-update, and `Client` is used by pointer so the `sync.Mutex` is never copied (vet agrees).
- Probe is bounded at 5s via `CommandContext`, failure reason is logged, and `unshare -r`'s uid-0-in-userns is documented in-line. `--`/no-`--fork` argv shape is now test-guarded so stdin piping survives.

That is real work. It still does not ship.

### Correctness Hawk — REJECT

`SecurityRequirements.Network` is a plain `bool` with `omitempty`. **Every existing manifest that omits `network` now means `network:false`, i.e. "isolate"** — and until this PR that field was purely declaratory (docs, line 72: "an inspectable declaration, not an OS sandbox guarantee"), so no plugin author had any reason to set `network: true`. On any Linux host with a working `unshare`, every already-installed plugin silently starts being launched inside a fresh netns. Two concrete breakages:

1. Plugins that make outbound calls (the whole point of a `source` plugin talking to GitHub/Jira) start failing with opaque network errors, with no manifest change and no release note.
2. A fresh netns comes up with **loopback down**. Plugins that talk to `127.0.0.1` (local sidecar, local model endpoint, local socket over TCP) break too — and that breaks even for plugins that legitimately declared no *external* network need.

This needs an explicit tri-state (`unset` = legacy permissive) or a schema-version-gated default, plus `ip link set lo up` inside the namespace if isolation is meant to keep loopback usable. As written the change is silently behavior-breaking for the installed base.

### Security Skeptic — REJECT

- **Fail-open is cached forever, silently-ish.** `sync.Once` memoizes a *failed* probe for the process lifetime. A transient failure at boot (load spike hitting the 5s timeout, a momentarily unavailable `unshare`) permanently downgrades the daemon to "no isolation" with only a `Warn` per invocation. There is also no strict/enforce mode: an operator who genuinely needs `network:false` honoured has no way to say "refuse to run rather than run unisolated". For a PR whose stated goal is network isolation, best-effort-only with no opt-in enforcement is an incomplete control.
- **The `size+mtime` cache makes the "re-verify before every invocation" claim stronger than the code.** An attacker who can write the plugin binary can also `touch -r` it; a same-size, mtime-preserving swap is never re-hashed. I accept this as *not fatal* — that attacker already owns the plugin root and can rewrite the manifest digest anyway, as your own comment concedes — but the comment currently promises per-invocation verification without stating the caveat. Given this PR's own standard ("honestly reported ... instead of silently pretending"), say so in the comment. Same for the inherent hash-then-exec TOCTOU window, which is unavoidable here but undocumented.

### Conventions Stickler — REJECT

Finding (10) is untouched and the docs are now **actively wrong**, not merely incomplete:
- `docs/plugins.md:72` still says `security` is "not an OS sandbox guarantee" and lines 143-144 still say "a malicious executable can still use the service account's ... network access". On Linux that is no longer true, and a plugin author reading the docs has no way to learn why their plugin lost the network.
- `checksum` is a new public manifest field and appears **nowhere** in `docs/plugins.md`'s manifest reference or in the example manifest. An undocumented security field is a field nobody uses.

`sandbox.go` is small, single-purpose, well-commented, and the argv-shape test is the right kind of test. No objection to the design.

### Pragmatist — APPROVE

Finding (8) — `ManifestSchemaVersion` still 1 while `DisallowUnknownFields` + exact-equality version check means an older Apiary reading a `checksum`-bearing manifest dies with `unknown field "checksum"` — I am **downgrading to non-blocking**. Bumping to 2 is strictly worse under the current `!=` check (it would reject every existing v1 manifest); the real fix is a supported-range check, which is its own change. It fails closed with a confusing message, on a pre-1.0 plugin surface. File it, don't block on it. Finding (6): the enforcement branch is still only exercised on hosts that can isolate, and there is still no stdin round-trip test through the wrapper — acceptable given CI is macOS, but note it.

### Verdict

Correctness and Security both reject → **REJECTED**.

**Blocking:**
1. Omitted `network` must not silently mean "isolate" for already-installed plugins; and bring loopback up inside the netns (or document that localhost dies).
2. Update `docs/plugins.md`: document `checksum`, and correct the now-false "declaration only / plugin still has network access" statements.

**Strongly recommended:** an opt-in strict mode that refuses to launch when `network:false` cannot be enforced; don't cache a *failed* probe for the process lifetime without any re-probe; document the `size+mtime` cache caveat and the hash→exec TOCTOU window on `integrityGuard`.

**Non-blocking follow-up:** schema-version range check so a `checksum` manifest degrades gracefully on older Apiary.

orlandoburli:
## Review council — round 3 — **PASS** (4/5 approve, no veto)

Re-reviewed at `7927f29` after the rescope to executable integrity only.

### Rescope verification
- `git diff main...HEAD` touches exactly 5 files; **no network-enforcement code remains** — no `unshare`/netns/isolation references anywhere under `internal/plugin/`, no dead helpers, no orphan imports. `SecurityRequirements.Network` is back to the untouched bare bool it was on `main`, so it is once again purely declaratory. The round-2 breaking change is genuinely gone, not merely defanged.
- `Invoke` execs exactly as before: same `exec.CommandContext(callCtx, c.executable)`, same `Dir`/`Env`/stdio/bounded buffers. The only delta is the `c.integrity.check` call before spawn.
- The PR description is honest: it states plainly that enforcement was dropped, why, that #292 stays open, and it self-documents the limits (tamper-evidence not authenticity, mtime+size-preserving swap defeats the cache, TOCTOU window, `DisallowUnknownFields` forward-compat wart). Scoping is legitimate — the residual claim is "integrity pinning", which the code delivers.

### Integrity code, end to end
- `normalizeChecksum`: trim → lowercase → strip prefix → trim → length → hex decode. The round-1 ordering bug is fixed and covered; blank means unpinned, non-blank malformed is an error (no silent unpinning).
- Wiring: `NewClient` verifies at construction; `Installed.Validate` calls `normalizeChecksum` so `apiary validate` catches a bad pin; `Invoke` re-verifies per call. Round-1 boot-only finding resolved.
- `integrityGuard`: mutex-guarded; unpinned fast path is correct; on failure `verified` is cleared so a bad state cannot be cached as good; the `stat` outside the lock can only store a *stale-older* mtime, which forces an extra re-hash rather than skipping one — conservative, not unsafe.
- Error paths: stat, open/read, and mismatch all return distinct, actionable messages; mismatch names both digests.

### Local checks
`gofmt -l internal/plugin/` clean · `go vet ./internal/plugin/` clean · `go test ./internal/plugin/` **ok**.

### Lenses
- **Correctness Hawk — APPROVE.** No new defect introduced by the removal; exec path byte-identical to pre-PR behavior apart from the check.
- **Security Skeptic — APPROVE.** Scope is narrower than the title once claimed, but the title/body now match what ships, and the limits are stated rather than oversold. Not a rollout gate per the operator-context review.
- **Conventions Stickler — REJECT (non-veto).** The file is still called `sandbox.go`/`sandbox_test.go` but contains no sandboxing — a leftover from the deleted scope. `integrity.go` is the honest name.
- **Simplicity Advocate — APPROVE.** ~117 lines, one concept, no feature-detection machinery left.
- **Pragmatist — APPROVE.** Ships a small real improvement; the hard design work is correctly deferred with the issue left open.

### Remaining (non-blocking)
1. Rename `sandbox.go` → `integrity.go` (and the test file) before merge.
2. Docs say the pin is verified "before the plugin is loaded" — precisely, it is at `NewClient`; harmless, but worth a word if you touch the file.
3. Track the deferred network-isolation design on #292 with the tri-state/`lo`/strict-mode requirements from round 2 recorded, so they are not rediscovered.

Issue #362, opened 2026-08-05
feat: monitoring sources Phase 2 — Dynatrace, webhook push, plugin-source bridge


orlandoburli:
Follow-up to #357, whose Phase 1 (Alertmanager poll source) shipped in v0.12.0 (#358) with test hardening in v0.12.1 (#361). These were explicitly scoped out of Phase 1 as separate PRs:

- **Dynatrace adapter**: `GET /api/v2/problems`, problem id as stable `SourceItem.ID`, `status: OPEN/RESOLVED` — same shape as the prometheus adapter, different client.
- **Generic webhook source**: wire `Adapter.WebhookHandler()` (currently unwired daemon-wide — no inbound HTTP listener exists). Enables Alertmanager push (`webhook_configs`) instead of polling and covers anything that can POST JSON (Loki ruler, Elastic Watcher). Needs listener + auth/HMAC design.
- **Plugin-source bridge**: the plugin protocol reserves `CapabilitySource` but the daemon only instantiates event-exporter plugins; a plugin-client → `source.Adapter` shim would let third parties ship monitoring sources out-of-process. Candidate alternative to in-tree Dynatrace.
- **`ack_via_silence` opt-in**: map `Acknowledge` to an Alertmanager silence.
- **Group-key dispatch**: dispatch per Alertmanager group instead of per alert (deferred in Phase 1 — groups have no stable fire-cycle epoch).
- **Resolved-while-running policy**: optional interrupt of a running investigation when its alert resolves (default remains: let it finish).

orlandoburli:
Dynatrace adapter shipped in #363: polls `GET /api/v2/problems`, `problemId` as the stable per-occurrence item ID, read-only shape mirroring the prometheus adapter (storm cap, flap dampener, capability lint). Remaining Phase 2 items (webhook push source, plugin-source bridge, `ack_via_silence`, group-key dispatch, resolved-while-running policy) still open here.

orlandoburli:
**Design decision (2026-08-05): the "Generic webhook source" item is dropped.** Apiary works **only with polling** — the daemon typically runs on a developer machine behind NAT, where an inbound webhook listener would require exposing a port or running a tunnel. Outbound polling works everywhere the daemon runs. PR #364 implemented it and was closed unmerged.

Remaining items on this issue: plugin-source bridge, `ack_via_silence`, group-key dispatch, resolved-while-running policy.

PR #364, opened 2026-08-06
feat(source): generic webhook source — push events into workflows via daemon HTTP listener (#362)

orlandoburli:
Implements the **Generic webhook source** item of #362: wires `source.Adapter.WebhookHandler()` daemon-wide so anything that can POST JSON (Alertmanager `webhook_configs`, Loki ruler, Elastic Watcher, CI scripts) pushes events straight into workflow trigger matching.

## Design (proposed here for review)

**Listener config** — one daemon-wide listener, `settings.webhook.listen: "127.0.0.1:8090"`. Started from `Dispatcher.StartServer` beside the queue protocol server; plain HTTP by design (TLS terminates at a reverse proxy on untrusted networks). `GET /healthz` for probes.

**Per-source path routing** — every source whose adapter returns a non-nil `WebhookHandler()` is mounted at `POST /webhook/{source-id}`. The daemon only routes by path; auth, parsing, and limits live in the adapter, so future push adapters (Dynatrace webhook, GitHub webhooks) reuse the same listener.

**Auth** (per-source `config`):
- `auth: bearer` (default) — `Authorization: Bearer <secret>`, constant-time compare. What Alertmanager's `http_config.authorization` can produce natively.
- `auth: hmac` — Stripe-style: `X-Apiary-Timestamp` + `X-Apiary-Signature: sha256=<hex HMAC-SHA256(secret, "<ts>.<body>")>`.
- `auth: none` — explicit opt-out, startup warning.

**Replay protection** (hmac mode) — the timestamp is bound into the signature and must be within `tolerance` (default 5m) of the daemon clock; valid signatures are cached and single-use inside the window. Outside the window the timestamp check alone rejects.

**Delivery → dispatch** — the handler parses (`format: generic` | `alertmanager`), applies `filters`, and enqueues into a bounded in-memory queue (`max_pending`, default 100 — 429 on overflow so the sender's retry re-delivers). The dispatcher drains it via the normal `Poll` path, so binding, routing, in-flight/once/active dedup, fan-out, and the durable queue all apply unchanged. A delivery signals a per-source wake channel so `pollLoop` runs immediately — `poll_interval` is only the fallback cadence.

**Identity / exactly-once** — generic events use their `id` field (or a body hash); Alertmanager alerts use `fingerprint:startsAt` exactly like the prometheus poll adapter, so a fleet can switch poll→push without changing dedup semantics. Re-deliveries still queued are dropped; re-deliveries after dispatch are shadowed by the persisted task/instance dedup.

**Validation** — new `config.SourcePushCapable` hook (probed in `cli/adapters.go` like the other capability hooks): a webhook-type source without `settings.webhook.listen` is a lint error. The adapter is read-only (no write capabilities), so the existing capability lint rejects `set_state`/labels/approvals/`wait_for: ci` against it.

## Changes
- `src/internal/source/webhook/` — adapter, auth (bearer/HMAC/replay), payload mapping (generic + Alertmanager), tests
- `src/internal/daemon/webhook_server.go` — listener + per-source mounting; `dispatcher.go` — wake channels, `pollLoop` wake case (additive; poll-only sources get a nil channel)
- `src/internal/config/` — `settings.webhook`, `SourcePushCapable` hook + lint, test
- Docs: `docs/webhook-source.md`, supported-integrations, mkdocs nav, `schema/apiary.json`, `.apiary/example-apiary-full.yaml`

## Testing
- `go test ./internal/...` green locally (no Go CI in this repo)
- Adapter tests: auth modes incl. replay/stale-timestamp/tampered-body, both formats, queue overflow/dedup/drain, filters, wake, read-only capability probes
- Daemon test: real TCP listener, routing to the right source, 404 for poll-only/unknown, healthz

Left for #362: webhook push for the dynatrace adapter, plugin-source bridge, `ack_via_silence`, group-key dispatch, resolved-while-running policy.

### DIFF
--- a/.apiary/example-apiary-full.yaml
+++ b/.apiary/example-apiary-full.yaml
@@ -170,6 +170,33 @@ sources:
       # match entity tags; raw criteria like displayId("P-123") pass through.
       labels: ["severity=availability", "managementZone=Prod"]
 
+  - id: alert-hooks
+    type: webhook
+    # Push-mode source: senders POST JSON to the daemon's webhook listener at
+    # /webhook/alert-hooks (requires settings.webhook.listen). Deliveries
+    # dispatch immediately; poll_interval is only the fallback cadence.
+    # Read-only source like prometheus/dynatrace: publish findings to a
+    # ticket source.
+    config:
+      # Shared secret for the selected auth mode (required unless auth: none).
+      secret: ${WEBHOOK_SECRET}
+      # bearer (default): Authorization: Bearer <secret> on every delivery.
+      # hmac: X-Apiary-Timestamp + X-Apiary-Signature (sha256=<hex HMAC over
+      #   "<timestamp>.<body>">), with replay protection inside `tolerance`.
+      # none: accept everything — only behind a trusted network boundary.
+      auth: bearer
+      # generic (default): one JSON object (or array / {"events": [...]}) with
+      #   optional id/title/description/labels/priority/state/url fields.
+      # alertmanager: the webhook_configs payload — firing alerts map exactly
+      #   like the prometheus poll source (fingerprint:startsAt identity).
+      format: alertmanager
+      # max_pending: 100        # pending-queue bound; overflow gets 429
+      # max_body_bytes: 1048576 # per-delivery body cap; larger gets 413
+      # tolerance: 5m           # hmac timestamp window
+    filters:
+      # Delivery-time filters: every label must be present on the event.
+      labels: ["severity=critical"]
+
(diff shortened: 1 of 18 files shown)


orlandoburli:
Closing unmerged by design decision: Apiary stays **polling-only**. An inbound webhook listener doesn't fit the product — the daemon typically runs on a developer machine behind NAT, where push would require exposing a port or a tunnel, while outbound polling works everywhere. The 'Generic webhook source' item of #362 is dropped.

Issue #367, opened 2026-08-06
feat(sdk): official plugin SDKs beyond Go


orlandoburli:
Follow-up to the plugin-source bridge (#362, PR #366). Protocol 1 is intentionally small (single-shot JSON over stdin/stdout), so plugins already work in any language — the docs ship **verified** Python, Rust, and Bash examples, each run against a real daemon. This issue tracks whether/when to provide *official* SDK packages so authors in other ecosystems get the envelope handling and typed source structs the Go SDK provides.

What an SDK must cover (from docs/plugin-sdk.md, 'Writing an SDK for your language'):
1. Read stdin to EOF, decode exactly one request; trailing data is an error
2. Verify `protocol == 1`; respond `unsupported_protocol` on mismatch
3. Echo `request_id` verbatim
4. Stdout purity: exactly one JSON response; all logging to stderr
5. Result/error dichotomy: exactly one of the two; `error` carries non-empty `code` + `message`
6. Exit 0 on delivered response
7. Typed mirrors of `sdk/plugin/source.go` (SourceItem: required stable `id`, RFC3339 timestamps, `key:value` labels)

Candidates, roughly by expected demand: **Python** (pip package, ~50 lines), **TypeScript/Node** (npm), **Rust** (crate). Open questions:
- Distribution: publish from this repo (monorepo folders like `sdk/python`) vs separate repos per language
- Versioning: SDK versions track the protocol version, not Apiary releases
- Conformance: a language-agnostic golden-file test kit (request/response fixtures) so every SDK validates against the same corpus — probably the first deliverable, since it also hardens the docs' hand-rolled examples

No urgency: the protocol-direct path is documented and proven; SDKs are ergonomics.

PR #448, opened 2026-08-27
feat(sdk): conformance kit for the plugin protocol, plus a Python SDK

orlandoburli:
Refs #367 — the conformance kit and the Python SDK. TypeScript/Node and Rust remain open, so this does not close the issue.

## The conformance kit (`sdk/conformance/`)

Protocol 1 is small enough to hand-roll in any language, which is the point — but "small" is not "obvious", and until now nothing checked that an implementation actually honoured the seven contract rules `docs/plugin-sdk.md` specifies. The kit is a **language-agnostic golden corpus**: JSON request fixtures plus the expectations each response must meet, driven against any plugin executable as a subprocess over stdin/stdout. The runner is `python3` stdlib only — no pip install, no virtualenv — so it costs the same in every ecosystem as it does here.

```bash
python3 sdk/conformance/run.py --config '{"path":"items.json"}' -- ./my-plugin
```

Ten cases, at least one per rule:

| Case | Rules | Checks |
|---|---|---|
| `poll-result-shape` | 1,3,4,5,6,7 | one JSON object, protocol 1, id echoed, one of result/error, exit 0; a result mirrors `SourcePollResult` — non-empty stable `id`, RFC3339 timestamps, string labels, no duplicate ids |
| `protocol-mismatch` | 2 | protocol 99 refused with error code exactly `unsupported_protocol` |
| `protocol-zero` | 2 | protocol 0 (a dropped field's zero value) refused the same way |
| `request-id-verbatim` | 3 | an id with spaces, punctuation and non-ASCII comes back byte for byte |
| `acknowledge-ok` / `write-result-ok` | 4,5,6 | write-back methods answer with a result, not an error |
| `unknown-method` | 5,6 | reported as an error with non-empty `code`+`message`, still exit 0 |
| `unknown-capability` *(should)* | 5,6 | advisory: a capability the plugin does not implement should be refused |
| `trailing-data` | 1 | a second object glued onto the stream must never yield a success result |
| `empty-stdin` | 1,4 | an empty stream carries no request, so no result may be invented |

Stdout purity (rule 4) is enforced on **every** case: each response is parsed strictly, and a stray `print`, banner or second line fails the case.

## Two real failures found

The issue calls the docs' hand-rolled examples "verified". The kit was the first chance to check that, and it found two violations — both fixed here:

- **The Go SDK broke rule 1.** `json.Decoder.Decode` stops at the first value and ignores the rest, so `ServeOne` answered a request with a second object glued onto the stream as if the exchange were healthy. It now treats trailing data as a transport error: nothing on stdout, handler never runs, exit 2. (`impact({target:"ServeOne", direction:"upstream"})` → LOW risk: one direct caller, `Main`.)
- **The Bash example in `docs/plugin-sdk.md` broke rule 2** — the one rule the docs spell out with a specific error code. It never read `protocol`, so it served a protocol-99 poll with real items. Fixed in the snippet.

Verbatim results after the fixes (`make conformance`):

```
conformance: go source-file (Go SDK)          10/10 cases passed
conformance: python source_file (Python SDK)  10/10 cases passed
conformance: in-tree source-bash              10/10 cases passed
conformance: in-tree source-node              10/10 cases passed
conformance: docs Python (plugins.md)         10/10 cases passed
conformance: docs Bash (plugin-sdk.md)        10/10 cases passed (1 warning)
conformance: docs Rust (plugin-sdk.md)        10/10 cases passed
CONFORMANCE OK — every checked plugin conforms to protocol 1
```

The docs' Bash snippet keeps one advisory warning: trimmed for the page, it does not switch on `capability`. That is a `should`, not one of the seven rules, and the prose now says so and points at `src/examples/plugins/source-bash` for the complete version.

`check-examples.sh` extracts the Python/Rust/Bash examples **verbatim from the Markdown** rather than duplicating them, so a documented snippet that drifts out of conformance fails CI instead of quietly misleading a reader. Extraction fails loudly when a section moves — a silently skipped example is worse than a broken build. Missing toolchains (no Node, no Rust) are skipped with a notice.

## The Python SDK (`sdk/python`)

Per the decision on distribution, non-Go SDKs are **monorepo folders** alongside the Go module, not separate repos.

- `apiary_plugin.protocol` — envelope handling: `main`, `serve_one`, `Request`, `PluginError`, `TransportError`, capability constants.
- `apiary_plugin.source` — typed mirrors of `sdk/plugin/source.go`, with `omitempty` semantics preserved, so a plugin written against either SDK puts the same bytes on the wire.
- Where Go returns `(result, *ResponseError)`, Python returns the result and raises `PluginError` — the same dichotomy, idiomatic in each language. `PluginError` rejects an empty `code` or `message` at construction, so rule 5 cannot be violated by accident.
- `pyproject.toml` makes it pip-installable from a checkout (`pip install ./sdk/python`, verified). **Not published to PyPI.**
- Stdlib only, `py.typed`, 18 unit tests on `python3 -m unittest`, and an example plugin that passes the corpus 10/10.

## Docs

- **Versioning and the protocol** is now a top-level section governing all SDKs — Go `sdk/v1.x`, Python `apiary-plugin` 1.x, both speaking protocol 1, both moving on a schedule independent of the daemon's `vX.Y.Z`. Consistent with the Go-specific section from #440, which now links to it.
- A **Python SDK** section with entry points and a complete example.
- **The conformance kit** section: how to run it, what a green run means, and a five-step recipe for a new-language SDK author (write the wrapper → 10/10 → mirror the typed structs → version against the protocol → add yourself to `check-examples.sh`).

## Wiring

- `make conformance`; `make test` also runs the Python unit tests.
- New `plugin-conformance` workflow on `sdk/**`, `src/examples/plugins/**`, `docs/plugins.md`, `docs/plugin-sdk.md` — Go tests, Python tests, then the kit.

## Verification

`cd src && go build ./... && go vet ./... && go test ./...`, the same for `sdk/`, `python3 -m unittest discover` in `sdk/python`, and `make conformance`. All clean.

## What remains open on #367

- **TypeScript/Node SDK** (npm) — deliberately not in this PR.
- **Rust SDK** (crate) — likewise.
- Publishing decisions: the Python package is in-tree only; nothing is on PyPI, and no npm/crates.io publication is set up.

Both remaining SDKs now have the corpus to validate against, which was the point of doing this first.

### DIFF
--- a/.github/workflows/plugin-conformance.yml
+++ b/.github/workflows/plugin-conformance.yml
@@ -1,0 +1,62 @@
+name: plugin-conformance
+
+# Runs the language-agnostic plugin protocol conformance kit against every
+# plugin implementation the repository ships — the Go and Python SDKs, the
+# in-tree Bash and Node example plugins, and the hand-rolled Python/Rust/Bash
+# examples extracted from the docs. A documented snippet that drifts out of
+# conformance fails here instead of quietly misleading a reader.
+on:
+  pull_request:
+    paths:
+      - "sdk/**"
+      - "src/examples/plugins/**"
+      - "docs/plugins.md"
+      - "docs/plugin-sdk.md"
+      - ".github/workflows/plugin-conformance.yml"
+  push:
+    branches: [main]
+    paths:
+      - "sdk/**"
+      - "src/examples/plugins/**"
+      - "docs/plugins.md"
+      - "docs/plugin-sdk.md"
+  workflow_dispatch: {}
+
+permissions:
+  contents: read
+
+jobs:
+  conformance:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+
+      - uses: actions/setup-go@v6
+        with:
+          go-version-file: src/go.mod
+          cache-dependency-path: src/go.sum
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.x"
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: "22"
+
+      # ubuntu-latest ships jq and a Rust toolchain; the kit skips anything
+      # that turns out to be missing rather than failing the run.
+      - name: Build the Node example plugin
+        working-directory: src/examples/plugins/source-node
+        run: npm install && npm run build
+
+      - name: Go SDK unit tests
+        working-directory: sdk
+        run: go test ./...
+
+      - name: Python SDK unit tests
+        working-directory: sdk/python
+        run: python3 -m unittest discover -s tests
+
+      - name: Conformance kit
+        run: sdk/conformance/check-examples.sh
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -7,6 +7,15 @@ version: 2
 
 project_name: apiary
 
+git:
+  # The SDK module is tagged `sdk/vX.Y.Z`, independently of the daemon's
+  # `vX.Y.Z`. Left alone, `git describe` picks whichever is nearest, and an
+  # SDK tag makes the derived version `sdk/v1.0.0-…` — a slash Docker rejects
+  # as an image tag ("invalid reference format"). Daemon releases are the only
+  # thing this config builds, so it only ever looks at `v*` tags.
+  ignore_tags:
+    - "sdk/*"
+
 before:
   hooks:
     - sh -c "cd src && go mod tidy"
--- a/Makefile
+++ b/Makefile
@@ -26,9 +26,18 @@ install: ## Install apiary to $$GOPATH/bin (makes it available on PATH)
 # ── test ───────────────────────────────────────────────────────────────────────
 
 .PHONY: test
-test: ## Run all tests (daemon module + SDK module)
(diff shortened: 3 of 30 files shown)


orlandoburli:
### Extra commit: `fix(release): keep SDK tags out of the daemon's derived version`

The `snapshot` check failed on the first push — **not** because of this PR. #440 introduced the `sdk/vX.Y.Z` tagging scheme, and `sdk/v1.0.0` was pushed 90 seconds *after* that PR's own CI had run, so nothing caught the consequence: `git describe` now resolves to `sdk/v1.0.0-…`, goreleaser derives that as the version, and Docker rejects the slash (`invalid tag … invalid reference format`). `release-check` has been red on main ever since for anything touching its trigger paths — this PR is just the first to hit it.

Since this config only builds daemon releases, it now ignores `sdk/*` tags. Verified locally: `goreleaser check` passes and the snapshot version is back to `0.18.2-SNAPSHOT-<sha>` instead of `sdk/v1.0.0-SNAPSHOT-<sha>`.

Unrelated to the SDK work, but it would have blocked every subsequent PR, so it is fixed here rather than filed.

Issue #472, opened 2026-09-01
Task list # column truncates a routine occurrence id, hiding the only thing that distinguishes rows


oburli:
The `#` column is sized for short references like `ERP-42` and `#412`. A routine occurrence id is 29 characters, so it truncates — and it truncates on the wrong side, cutting off the only part that tells two rows apart.

## What it looks like

With [`dev.apiary.routines`](https://github.com/orlandoburli/apiary-routines) emitting a work item per cron occurrence, the reference is `<routine id>@<instant>`. Two occurrences of a 5-minute routine:

```
$ sqlite3 apiary.db "select b.source_item_number, t.title from source_bindings b
                     join internal_tasks t on t.id = b.task_id where b.source_id='routines'"
pr-watch@2026-09-01T22:00:00Z  PR watch
pr-watch@2026-09-01T22:05:00Z  PR watch
```

and how they render (`apiary instances`, CELL column — the same sizing assumption as the dashboard's `#`):

```
ID                          WORKFLOW                CELL          STATE     STARTED     DURATION
wf_1788300321365370000_5    pr-sweep-watch          pr-watch@20…  running   01:43 ago   01:43
wf_1788300021402022000_1    pr-sweep-watch          pr-watch@20…  done      06:43 ago   01:40
```

**The two rows are now indistinguishable.** The title is the routine's configured `title` — identical for every occurrence by design — so the reference is the *only* thing that separates the 22:00 run from the 22:05 run, and it is exactly what got cut. Truncation kept `pr-watch@20`, which every occurrence of this routine shares forever, and dropped the timestamp, which is the entire distinguishing content.

Reference widths in play, from one hive running three source kinds:

| source | example | length |
|---|---|---|
| `routines` (plugin) | `pr-watch@2026-09-01T22:00:00Z` | 29 |
| jira | `CDT-109` | 7 |
| github | `#49` | 3 |

`docs/dashboard.md` describes the column as:

> The `#` column shows each task's human reference (e.g. `ERP-42`) so you can match a row to the work item in Plane.

which is precisely the assumption — references are short and Plane/GitHub-shaped. That held until a source started minting synthetic references, and the plugin SDK means more of those are coming.

## Suggested fix, roughly in order of preference

1. **Size the column to its content** — widest visible reference, capped at some sane maximum, with the slack taken from whichever column has it (`TASK` is usually the longest and the most compressible, since a title truncated mid-sentence is still readable).
[...]

PR #483, opened 2026-09-03
feat: filter task list to ticket-bound work (--tickets-only)

### LINKED ISSUE #475: Task list has no filter to separate ticket-bound work from scheduled/routine runs
Split out of #474 (the task-list half; the approval-visibility half is tracked separately).

## Problem

The task/instance list is dominated by scheduled and manually-triggered routine runs (e.g. `pr-sweep-watch`, `pr-sweep-dispatch`) that have no underlying Jira ticket or GitHub issue — they're plumbing, not work. These fill up the list and bury the actual ticket-driven work (`jira-implement`, `pr-checkup` dispatched against a real ticket, etc.), making it hard to see what's actually going on.

## Ask

Separate "real" tasks (bound to a Jira ticket or GitHub PR/issue via `source_bindings`) from scheduled/routine-triggered runs (no ticket binding) in the task list — either as distinct views/tabs, or a clear visual/filter split, so routine noise doesn't drown out ticket work.

## Why this matters

With [`dev.apiary.routines`](https://github.com/orlandoburli/apiary-routines) minting a work item per cron occurrence (see also #472), routine-run volume in the list only grows. Today, finding real ticket work means going around the dashboard entirely — CLI flags or direct sqlite queries — instead of just looking at the task list.

## Suggested approach

A source-kind or binding-presence filter (`source_bindings` row exists vs. not) is enough to split the two populations; it doesn't need per-source special-casing since any future plugin-sourced routine will look the same (no ticket binding).

orlandoburli:
## Summary

- `apiary instances` gains a `--tickets-only` flag that excludes instances with no real ticket behind them — scheduled routine runs and other plugin-sourced work items — keeping only rows bound to a source item from a ticket-tracker source (jira, github, plane, ...).
- The dashboard's Tasks tab gets the same filter as a `Shift+T` toggle on the list view, shown in the title (`[tickets-only]`) and footer hint.
- The distinguishing signal is structural: `config.IsTicketSourceType(sourceType)` treats any source config type other than `"plugin"` (the generic plugin-bridge kind used by the routines source and future plugin-sourced monitoring/scheduler adapters) as a ticket tracker. A task/instance counts as ticket-bound when it has a `source_bindings` row pointing at such a source — no per-source-id special-casing, so a future plugin-sourced routine is excluded automatically.
- `ListWorkflowInstanceViews` now takes an optional ticket-source-id allow-list so the filter is applied in the SQL query itself (`WHERE wi.source_id IN (...)`), keeping `--limit` semantics correct instead of trimming an already-limited result set. `nil` (the default) is unfiltered — behavior for every existing caller is unchanged.
- Both defaults are off, so today's unfiltered output (CLI and dashboard) is unchanged; this is purely additive.

Does **not** touch #472's column-width rendering for routine references — this issue is about which rows show up, not how a row renders.

## Test plan

- [x] `cd src && go build ./... && go vet ./...`
- [x] `cd src && go test ./internal/cli/... ./internal/dashboard/... ./internal/db/... ./internal/config/...`
- [x] `cd src && go test ./...` (full suite, no regressions)
- [x] New tests: `internal/config/ticket_source_test.go` (`IsTicketSourceType`, `Config.TicketSourceIDs`), `internal/db/instance_views_ticket_filter_test.go` (nil vs. populated vs. empty allow-list), `internal/dashboard/tickets_only_test.go` (`hasTicket` classification + `Shift+T` toggle + `filteredTasks`)

Closes #475

### DIFF
--- a/src/internal/cli/instances.go
+++ b/src/internal/cli/instances.go
@@ -30,11 +30,12 @@ var (
 
 func newInstancesCmd() *cobra.Command {
 	var (
-		workflow string
-		state    string
-		limit    int
-		asJSON   bool
-		cancel   bool
+		workflow    string
+		state       string
+		limit       int
+		asJSON      bool
+		cancel      bool
+		ticketsOnly bool
 	)
 
 	cmd := &cobra.Command{
@@ -45,7 +46,10 @@ func newInstancesCmd() *cobra.Command {
 			"and the instance is marked interrupted, without touching the source item's\n" +
 			"labels or state (unlike `apiary restart`, which acts on the whole cell).\n" +
 			"Queued or leased dispatch jobs for the same task and workflow are cancelled\n" +
-			"with it. A cancelled instance can be continued later with `apiary resume`.",
+			"with it. A cancelled instance can be continued later with `apiary resume`.\n\n" +
+			"--tickets-only hides instances with no real ticket behind them — scheduled\n" +
+			"routine runs and other plugin-sourced work items — keeping only instances\n" +
+			"bound to a source item from a ticket-tracker source (jira, github, plane, ...).",
 		Args: cobra.MaximumNArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if cancel {
@@ -57,7 +61,7 @@ func newInstancesCmd() *cobra.Command {
 			if len(args) == 1 {
 				return showInstance(args[0], asJSON)
 			}
-			return listInstances(workflow, state, limit, asJSON)
+			return listInstances(workflow, state, limit, asJSON, ticketsOnly)
 		},
 	}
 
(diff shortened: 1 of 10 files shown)


orlandoburli:
Pushed a fix for a real gap in the ticket-vs-routine classification.

`IsTicketSourceType` was a block-list (`sourceType != "plugin"`), but `dynatrace` and `prometheus` register under their own distinct types — not `"plugin"` (see `source.Register("dynatrace", ...)` / `source.Register("prometheus", ...)`). So a Dynatrace or Prometheus alert, which mints a synthetic work item with no real ticket behind it exactly like a routine occurrence, was being counted as ticket-bound and would still show up under `--tickets-only`.

Switched to an explicit allow-list of the registered ticket-tracker types (`jira`, `github`, `plane`). This trades away "a future ticket-tracker type needs zero code changes" for "a future non-ticket adapter is excluded by default instead of silently included" — the latter is the safer failure mode for a filter whose whole point is hiding plumbing. Added `dynatrace`/`prometheus` cases to the test table so this doesn't regress silently again.

`go build ./... && go vet ./... && go test ./...` all pass after the change.

